PR #150, opened 2025-02-27
feat: dark/light mode toggle in settings

JiyaGupta-cs:
### Description
This PR introduces a **dark/light mode** toggle feature, allowing users to switch between themes dynamically. The selected theme remains **persistent** even after refreshing the website.

Key Features:
- Implemented dark/light mode toggle using DaisyUI.
- Users can switch between various light mode themes provided by DaisyUI.
- Theme preference is stored persistently using cookie.
- Ensured smooth UI updates on theme change.

### Additional Context
The logo also needs to be changed for the light mode and the theme change option need to be styled as the dropdown like other similar fields.
If you prefer to change the light mode colors, it can be selected from any theme of DaisyUI: [https://daisyui.com/docs/themes](https://daisyui.com/docs/themes). Let me know if you'd like a specific theme to be set as the default.

Looking forward to your feedback!

### Screenshot

[Screencast from 2025-02-28 04-38-36.webm](https://github.com/user-attachments/assets/5492c5f4-1667-4899-858d-d7c310d6c786)

### Fixes bug(s)
Closes the dark/light mode feature on #2

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" data-theme="">
 	<head>
 		<meta charset="utf-8" />
 		<link rel="icon" href="%sveltekit.assets%/favicon.png" />
--- a/src/hooks.server.ts
+++ b/src/hooks.server.ts
@@ -1,0 +1,21 @@
+import type { Handle } from "@sveltejs/kit";
+
+export const handle = (async ({ event, resolve }) => {
+    let theme: string | null = null;
+    const url = new URL(event.request.url); 
+    const newTheme = url.searchParams.get('theme');
+    const cookieTheme = event.cookies.get('colortheme');
+    if (newTheme) {
+        theme = newTheme;
+    } else if (cookieTheme) {
+        theme = cookieTheme;
+    }
+    if (theme) {
+        return await resolve(event, {
+            transformPageChunk: ({ html }) =>
+                html.replace('data-theme=""', `data-theme="${theme}"`),
+        });
+    }
+    return await resolve(event);
+}) satisfies Handle;
+
--- a/src/routes/+page.server.ts
+++ b/src/routes/+page.server.ts
@@ -1,0 +1,13 @@
+import type { Actions } from "@sveltejs/kit";
+
+export const actions: Actions = {
+	setTheme: async ({ url, cookies }) => {
+		const theme = url.searchParams.get('theme');
+		if(theme){
+			cookies.set('colortheme', theme, {
+				path: '/',
+				maxAge: 60 * 60 * 24 * 365,
+			});
+		}
+	}
+};
--- a/src/routes/settings/+page.svelte
+++ b/src/routes/settings/+page.svelte
@@ -5,6 +5,8 @@
 	import Heading from './Heading.svelte';
 	import { FolksonomyApi } from '$lib/api/folksonomy';
 	import { t } from '$lib/translations';
+	import type { SubmitFunction } from '@sveltejs/kit';
+	import { enhance } from '$app/forms';
 
 	interface Props {
 		data: PageData;
@@ -19,6 +21,13 @@
 
 		new FolksonomyApi(fetch).login(username, password);
 	}
+
+	const submitUpdateTheme: SubmitFunction = ({action})=>{
+		const theme = action.searchParams.get('theme');
+		if(theme){
+			document.documentElement.setAttribute('data-theme', theme);
+		}
+	} 
 </script>
 
 <div class="mx-auto my-8 grid grid-cols-[1fr,max-content] items-center gap-x-8 gap-y-2">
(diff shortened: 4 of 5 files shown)


VaiTon:
Hello @JiyaGupta-cs and thanks for the PR. I don't think storing the theme in cookies is a pretty way to do so. We already have a mechanism to store user preferences in local storage (`src/lib/settings.ts`). Please use that to store the user theme.

If you'd like a reference, look at this code: https://github.com/cartabinaria/dynamik/blob/main/src/lib/settings.ts

Sharma-Ji-21:
@VaiTon  This issue was assigned to Jiya, but there hasn’t been any progress in the last few days. I’d like to work on it if possible. Also, I will be working on the issues already assigned to me, as I couldn’t contribute earlier due to personal reasons. Let me know if I can take this one up!

JiyaGupta-cs:
@VaiTon I am working on this issue.
It was working fine before: 
[a2d8fb3d783c1055f2564ae5c3101567dd7ff316](https://github.com/openfoodfacts/openfoodfacts-explorer/commit/a2d8fb3d783c1055f2564ae5c3101567dd7ff316), 
[38b503ebd711de4ae423e9e96aaf46d34f4c2bb1](https://github.com/openfoodfacts/openfoodfacts-explorer/commit/38b503ebd711de4ae423e9e96aaf46d34f4c2bb1)
[875456b6b9621afaed0f000b91dc92c6a5a2020b](https://github.com/openfoodfacts/openfoodfacts-explorer/commit/875456b6b9621afaed0f000b91dc92c6a5a2020b)
I am looking into it

JiyaGupta-cs:
[Screencast from 2025-03-05 03-50-42.webm](https://github.com/user-attachments/assets/8d5fa146-8222-4a47-bf02-093f28a04109)


I have updated the code taking inferences from the reference provided. Thanks !

src/lib/settings.ts
```diff
@@ -12,3 +12,19 @@ export const preferences = persisted('preferences', {
...
+	'light',
+	'dark',
+] as const;
+type Theme = (typeof themes)[number];
+
+export interface Settings {
+	theme: Theme;
+}
+
+const settings = persisted<Settings>('settings', {
```
VaiTon, line 26: Why another persisted? Can't we use the `preferences` on line 3?

src/lib/ui/Logo.svelte
```diff
@@ -1,10 +1,19 @@
...
+	let imageUrl = 'https://static.openfoodfacts.org/images/logos/off-logo-horizontal-dark.svg';
+
+	onMount(() => {
+		settings.subscribe(($settings) => {
+			imageUrl =
+				$settings.theme === 'dark'
+					? 'https://static.openfoodfacts.org/images/logos/off-logo-horizontal-dark.svg'
+					: 'https://static.openfoodfacts.org/images/logos/off-logo-horizontal-light.svg';
+		});
+	});
```
VaiTon, line 14: Can't we do this with a `$derived`?

VaiTon (changes requested):

src/lib/settings.ts
JiyaGupta-cs: Yes, you are right. I have corrected that

src/lib/ui/Logo.svelte
JiyaGupta-cs: I have updated that. Thanks

VaiTon:
> Users can switch between various light mode themes provided by DaisyUI.

Mh, but with this PR we're just making changing theme between light and dark mode manual. I don't see any new theme added.

Now the question is, why should we do something like this?

JiyaGupta-cs:
> > Users can switch between various light mode themes provided by DaisyUI.
> 
> Mh, but with this PR we're just making changing theme between light and dark mode manual. I don't see any new theme added.
> 
> Now the question is, why should we do something like this?

Yes, you are right we are just switching between the dark and light mode. I will update the PR description

VaiTon:
> Yes, you are right we are just switching between the dark and light mode. I have updated the PR description now

The PR description is not the problem. The question is: why should the user NOT use the automatic color switching and instead fix the theme? At least we should provide a way to use the auto color switching.

JiyaGupta-cs:
@VaiTon
using ```prefersdark: true;``` uses auto color switching and changes it according to the user's browser preference but using that the user is unable to switch between the light and dark mode manually.
I have tried to imitate the same behaviour as of [https://github.com/cartabinaria/dynamik](https://github.com/cartabinaria/dynamik/)
If you want I can change it to the auto color switching but then the user will not be able to switch the mode manually.

JiyaGupta-cs:
@VaiTon could you please guide more about it

VaiTon:
@JiyaGupta-cs if we want to make the user able to select a light / dark mode and keep it, we should at least add a "system" / "auto" mode that follows OS defaults

JiyaGupta-cs:
> @JiyaGupta-cs if we want to make the user able to select a light / dark mode and keep it, we should at least add a "system" / "auto" mode that follows OS defaults

The problem that is occuring is ff ```--prefersdark``` is used as mentioned in [https://daisyui.com/docs/themes/](https://daisyui.com/docs/themes/) to detect the current theme of the system then the user is unable to switch between the modes

VaiTon:
Then I'm sorry @JiyaGupta-cs but I don't think that is worth the change. Not a lot of users would want to fix a theme anyway

Issue #169, opened 2025-03-01
Implement tag fields editing (categories, labels, origins, stores, traceability codes, countries…)


VaiTon:
- [x] Categories
- [x] Labels
- [x] Origins
- [x] Stores
- [ ] Traceability codes
- [x] Countries

lakshaydewan:
@VaiTon Would love to be assigned.

KarruHarin:
@VaiTon I would be happy to receive feedback on my pr ,any changes to be made!!

VaiTon:
@KarruHarin, thank you for your PR but you should check if anyone else is working on an issue before working on it. In this case, @lakshaydewan had already taken the job. Tomorrow I'll check the PR.

KarruHarin:
@VaiTon my bad , I will make sure to check it from next time !

VaiTon:
@lakshaydewan are you ok with #184 ? Do you want to do it nonetheless?

lakshaydewan:
If @KarruHarin has raised a PR already maybe he should go on with this I'll pick up something else.

VaiTon:
@lakshaydewan , I think there is something left in this issue. If you have some time please check and report back. Thanks!

lakshaydewan:
@VaiTon sure; tell me what is needed.

KarruHarin:
@VaiTon I will be taking it ,nd submit pr by today night.
if @lakshaydewan is okay , otherwise I'm okay even if @lakshaydewan  takes it.

lakshaydewan:
@KarruHarin u go ahead and:)

PR #182, opened 2025-03-01
Product Editing testing with  store

KarruHarin:
### What
I added store to the product Editing to get your opinion ,if your okay with my procedure I will continue to add other tag fields editing (origins,tracebility codes etc..)

### Screenshot
Screenshot 2025-03-02 020612
Screenshot 2025-03-02 020447

Key Features :
#169
Added tag fields editing(store)

Additional Info :
Will add other editing field if my methods correct

Looking forward to your feedback!


VaiTon:
@KarruHarin please rebase on main and then run `pnpm format`.

Issue #176, opened 2025-03-01
Allow viewing multilingual fields (ingredients, product name)


VaiTon:

Abbas-Askari:
@VaiTon I'll work on this.

VaiTon:
Okay, I'm assigning you

Abbas-Askari:
@VaiTon I think this can be closed.

Issue #125, opened 2024-10-30
Implement product editing


teolemon:
### What

```[tasklist]
### Tasks
- [ ] Implement basic details editing 
- [ ] Implement product name
- [ ] Implement brand
- [ ] Implement quantity 
- [ ] Implement tag fields editing (categories, labels, origins, stores, traceability codes, countries…)
- [ ] Implement ingredients editing
- [ ] Implement packaging editing
- [ ] Implement producer website editing
- [ ] Implement Folksonomy Engine editing
- [ ] https://github.com/openfoodfacts/openfoodfacts-explorer/issues/114
- [ ] Implement a photo manager
- [ ] Add support for Price edition
```

teolemon:
- [ ] Allow answering the Robotoff questions
![Image](https://github.com/user-attachments/assets/8630dccb-e933-4d00-8ca5-6c55d30dd53a)
- [ ] Allow logging to Open Prices with your main Open Food Facts account (allowing logging with a different account is an idea that should be kept, for people willing to have maximum privacy with an account with their name)
![Image](https://github.com/user-attachments/assets/2d80f3ba-5a9b-4645-bef2-6b3eaa56e559)

Sharma-Ji-21:
@teolemon i would like to work on this issue, could u pls assign this to me

KarruHarin:
@teolemon I tried to clone the repo to set up dev environment , i was able to set it up but was unable to move on to individual product page , as it have its Base_url in .env file of the project .Can I pls get guidence on how to completely set it up ?

VaiTon:
@JiyaGupta-cs sure you can, but please create a PR for a specific bullet point. It would be better to first ask about a specific bullet point so that I can create an issue about it.

@Sharma-Ji-21 same thing. This is an "Epic". It won't be assigned to anyone. It's just to track the missing features. Please ask for a specific bullet point and I can assign you.

@KarruHarin please message me on Slack.

Abbas-Askari:
@VaiTon Can I work on `Implement basic details editing`, I'll add the form feilds in the  `http://localhost:5173/products/[barcode]/edit` page and connect it to the api.

PR #185, opened 2025-03-02
feat: Add editing support for ingredients and name in all available languages

Abbas-Askari:
### What
for each language in the product.languages_codes, show a tab in productnames and indegredients.

### Screenshot
image
image
Only tabs for available existing languages are show:
image

### Fixes bug(s)
- #176 

### Part of 
- #125

### DIFF
--- a/src/lib/api/product.ts
+++ b/src/lib/api/product.ts
@@ -35,6 +35,20 @@ export class ProductsApi {
 		if (!username || !password) throw new Error('No username or password set');
 
 		const languageCodes = Object.keys(product.languages_codes);
+		const productNames = languageCodes.reduce(
+			(acc, lang) => {
+				acc[`product_name_${lang}`] = product[`product_name_${lang}`];
+				return acc;
+			},
+			{} as Record<string, string>
+		);
+		const ingredientsTexts = languageCodes.reduce(
+			(acc, lang) => {
+				acc[`ingredients_text_${lang}`] = product[`ingredients_text_${lang}`];
+				return acc;
+			},
+			{} as Record<string, string>
+		);
 
 		const body = formData({
 			code: product.code,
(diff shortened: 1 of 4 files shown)


VaiTon:
I don't think creating a field for each language is a good approach. Maybe index signatures are a better one.

https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures

Abbas-Askari:
@VaiTon Its done. I have generalized the use of lanugage specific keys.

src/lib/api/product.ts
```diff
@@ -42,7 +44,25 @@ export class ProductsApi {
...
+			comment: product.comment ?? '',
+
+			product_name: product.product_name,
+			...languageCodes.reduce(
+				(acc, lang) => {
+					acc[`product_name_${lang}`] = product[`product_name_${lang}`];
+					return acc;
+				},
+				{} as Record<string, string>
+			),
```
VaiTon, line 56: Very elegant but a little difficult to read. It would be better to extract a variable here.

src/lib/api/product.ts
```diff
@@ -42,7 +44,25 @@ export class ProductsApi {
...
+
+			ingredients_text: product.ingredients_text,
+			...languageCodes.reduce(
+				(acc, lang) => {
+					acc[`ingredients_text_${lang}`] = product[`ingredients_text_${lang}`];
+					return acc;
+				},
+				{} as Record<string, string>
+			)
```
VaiTon, line 65: Same as before

src/routes/products/[barcode]/edit/+page.svelte
```diff
@@ -46,23 +51,52 @@
 		}
 	}
 
+	function getIngredientsImage(language: string) {
+		let baseUrl = 'https://images.openfoodfacts.org/images/products/';
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/+page.svelte:55]: Make this a const and put it in a `lib/` file

VaiTon (changes requested):

Abbas-Askari:
Sure, I'll make the necessary changes soon.

src/routes/products/[barcode]/edit/+page.svelte
```diff
@@ -93,18 +127,32 @@
...
+					<img src={getIngredientsImage(code)} alt="Ingredients" class="mb-4" />
+				{:else}
+					<p class="alert alert-warning mb-4">No ingredients image</p>
+				{/if}
+				<label for="">Ingredients list ({getLanguage(code)})</label>
+				<div class="form-control mb-4">
+					<textarea
+						class="textarea textarea-bordered h-40 w-full"
+						bind:value={
+							$productStore[('ingredients_text_' + code) as keyof typeof data.state.product]
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/+page.svelte:150]: Can we make this line more comprehensible?

Comment (Abbas-Askari) [reply src/routes/products/[barcode]/edit/+page.svelte]: I have updated it to
```jsx
bind:value={$productStore[`ingredients_text_${code}`]}
```
I hope this is fine, Otherwise ```$productStore[`ingredients_text_${code}`]``` will need be extracted as a const.

VaiTon (approved):

PR #186, opened 2025-03-02
fix: Broken pagination

Abbas-Askari:
### What
Fix the broken pagination.
There are only 7 pages for "abc" query, but the pagination section show 50 pages.
Actually, the previous code assumes `page_count` is the total number of pages, but it is actually `Number of products in this page` as obvious from the [API docs](https://openfoodfacts.github.io/openfoodfacts-server/api/ref-v2/#get-/api/v2/search):
image
The Api return no feild indicting total amount of pages, So this PR calculates the feild `total_pages` and uses it instead of `page_count` :)

### Screenshot
Before:
image
After:
image

### DIFF
--- a/src/routes/products/search/+page.svelte
+++ b/src/routes/products/search/+page.svelte
@@ -41,15 +41,15 @@
 
 			<button class="btn join-item btn-active">{result.page}</button>
 
-			{#if result.page_count > result.page + 1}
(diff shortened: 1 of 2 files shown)


VaiTon (approved):

VaiTon:
Thank you! :tada:

PR #198, opened 2025-03-03
Fix: Prevent Dropdown Overlap in Tag Selection Input

### LINKED ISSUE #192: UI Issue: Category Tags Overlapping Input Field
### Description
The category tags are overlapping the input field, making it difficult to see and type new entries properly. This affects usability and could cause frustration for users trying to add or edit categories.

### Steps to Reproduce

1. Start typing a new category in the input field.
2. If there are many existing tags, they overlap the input field instead of wrapping properly.
3. The dropdown suggestions also appear in an inconsistent manner.

### Expected Behavior

1. The input field should remain clearly visible.
2. Tags should wrap properly inside the container without overflowing into the input area.
3. The dropdown suggestions should not overlap with the existing tags.

### Screenshots

1. 
Image

2. 
Image

SyedImtiyaz-1:
### Description:
This PR fixes the dropdown overlap issue in the Tag selection input field by restructuring the dropdown placement. The dropdown was previously constrained within the input container, causing it to be overlapped by selected tags and other elements.

### Fix Implemented:
✅ Moved the dropdown outside the input wrapper to ensure it stays below the field.
✅ Applied absolute top-full left-0 w-full to properly align the dropdown.
✅ Added z-[100] to ensure it appears above other elements.
✅ Used border border-gray-300 rounded-md for a better visual distinction.
✅ Improved onblur behavior to prevent unintended closing.

### Screenshots 
1. 
Screenshot from 2025-03-03 21-42-58

2.
Screenshot from 2025-03-03 21-43-27

#### Steps to Reproduce the Fix:
1. Click inside the "Brands" input field.
2. Start typing to trigger the dropdown suggestions.
3. Observe that the dropdown now stays below the input field instead of overlapping.
4. Select a suggestion, and the dropdown closes as expected.

Related Issues:
- Closes #192

### DIFF
--- a/src/routes/products/[barcode]/edit/Tags.svelte
+++ b/src/routes/products/[barcode]/edit/Tags.svelte
@@ -16,6 +16,7 @@
 	);
 
 	let newValue = $state('');
+	let showDropdown = $state(false);
 
 	let filteredAutocomplete = $derived(
 		newValue.length < 3 ? [] : autoCompleteFuse.search(newValue).slice(0, 10)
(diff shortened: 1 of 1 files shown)


src/routes/products/[barcode]/edit/Tags.svelte
```diff
@@ -16,6 +16,7 @@
 	);
 
 	let newValue = $state('');
+	let showDropdown = $state(false);
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/Tags.svelte:19]: Why do we need another state variable? Is the `filteredAutocomplete.length > 0` check not enough?

Comment (SyedImtiyaz-1) [reply src/routes/products/[barcode]/edit/Tags.svelte:19]: I considered removing `showDropdown`, but I kept it because it provides better control over when the dropdown appears. Using only `filteredAutocomplete.length > 0` would make it disappear immediately when there are no matches, which can cause flickering. With `showDropdown`, I can ensure a smoother UX, keeping it open while typing and only closing it when focus is lost. Let me know if you have any suggestions!

src/routes/products/[barcode]/edit/Tags.svelte
```diff
@@ -45,44 +46,48 @@
...
-								{key}
-							</button>
-						</li>
-					{/each}
-				</ul>
-			</div>
-		{/if}
 	</div>
+
+	<!-- Dropdown Suggestions (placed outside the input wrapper) -->
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/Tags.svelte:68]:
```suggestion
	<!-- Dropdown Suggestions -->
```

src/routes/products/[barcode]/edit/Tags.svelte
```diff
@@ -45,44 +46,48 @@
...
+				{tag}
+				<button class="ml-2 text-xl" onclick={removeTag(tag)}>×</button>
+			</span>
+		{/each}
 		<input
 			type="text"
-			class="w-full bg-transparent outline-hidden"
+			class="z-10 w-full bg-transparent outline-none"
+			onfocus={() => (showDropdown = true)}
+			onblur={() => setTimeout(() => (showDropdown = false), 200)}
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/Tags.svelte:62]: This is not ok. When I use tab to focus the list it immediately disappears

Comment (SyedImtiyaz-1) [reply src/routes/products/[barcode]/edit/Tags.svelte:62]: Okay, Got it. I'll make the changes.

src/routes/products/[barcode]/edit/Tags.svelte
Comment (VaiTon) [reply src/routes/products/[barcode]/edit/Tags.svelte:19]: Okay

SyedImtiyaz-1:
@VaiTon , creating a new PR fixing all the check !

Issue #171, opened 2025-03-01
Implement Folksonomy Engine editing


VaiTon:

lakshaydewan:
@VaiTon  Would love to work on this.

heymitali:
@VaiTon can i work on this, if this task is available ?

swastikCommits:
Isn't this fixed in #212 ?

PR #212, opened 2025-03-05
Fix: fixed the token error in login

lakshaydewan:
What
Fixes the token not being correctly parsed in the login method.

Fixes bug(s)
#171


lakshaydewan:
@VaiTon; if we login with the correct username password from the official OFF account, the generated token is being correctly setUp, plus the tags viewing, adding and deleting are also working. Waitin for review.

VaiTon:
Thank you! :tada:

Issue #224, opened 2025-03-07
Option to scan barcodes thorugh camera disappears on smaller screens


Prathamesh0901:
### What

When website is opened on wider screens (e.g. laptops) we can see the option to scan the barcode through the camera, but on smaller screens (e.g. mobile) the option disappears.

![Image](https://github.com/user-attachments/assets/0c1c7ab8-ee24-476a-bf9e-346f701f393e)

### Steps to reproduce the behavior:
1. Open website on a mobile.

### Expected behavior

The option to scan the barcode should be visible on all devices.

Prathamesh0901:
I am open to work on this issue.

PR #229, opened 2025-03-07
fix: visibility of barcode scan option on smaller screen

Prathamesh0901:
### What
- The option to scan barcodes through smaller screen devices was not visible previously.

### Screenshot

image

### Fixes bug(s)
- #224

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -131,25 +131,35 @@
 		</div>
 	</div>
 	{#if searchActive}
-		<div class="join -mt-8 w-full">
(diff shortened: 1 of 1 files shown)


VaiTon:
Instead of displaying it only when the search bar is open, I think it would be better to display it near the search icon on the top-right of the screen.

Prathamesh0901:
> Instead of displaying it only when the search bar is open, I think it would be better to display it near the search icon on the top-right of the screen.

ok, then I will place it in between search icon and options icon.

PR #235, opened 2025-03-08
Replace Camera Icon with Barcode Icon for QR Scanner Button

RISHIK92:
The QR scanner button previously displayed a camera icon (mdi--camera), which was not the most intuitive choice for barcode scanning. This PR updates the button to use a barcode icon (mdi--barcode-scan) for better clarity and user experience.

Changes Made:
Replaced <span class="icon-[mdi--camera]"></span> with <span class="icon-[mdi--barcode-scan]"></span>
Ensured the new icon aligns correctly with the button style
Why this change?
The barcode icon better represents the functionality of scanning a product barcode.
Enhances usability by making the button's purpose immediately clear.

Testing:
Manually verified that the barcode icon displays correctly in the UI.
Ensured no other UI elements were affected.


src/routes/qr/+page.svelte
```diff
@@ -2,8 +2,7 @@
 	import { onDestroy, onMount } from 'svelte';
 	import { Html5Qrcode, Html5QrcodeSupportedFormats } from 'html5-qrcode';
 
-	let error: string | null = $state(null);
```
VaiTon, line 5: Why did you remove the `$state`?

src/routes/qr/+page.svelte
```diff
@@ -18,10 +17,15 @@
 			verbose: false
 		});
 
+		const getQrBoxSize = () => {
+			const screenWidth = window.innerWidth;
+			return screenWidth < 640 ? { width: 250, height: 250 } : { width: 400, height: 250 };
+		};
```
VaiTon, line 23: Use a `function` declaration

src/routes/qr/+page.svelte
```diff
@@ -42,19 +46,18 @@
...
 	onDestroy(() => {
-		// Stop the scanner when the component is destroyed
-		if (html5QrCode != null) {
+		if (html5QrCode) {
 			html5QrCode.stop();
 		}
 	});
 </script>
 
-{#if error != null}
```
VaiTon, line 52: Keep it `!= null`

VaiTon (changes requested):

src/routes/qr/+page.svelte
```diff
@@ -42,19 +46,18 @@
...
-	<div class="flex items-center justify-center">
-		<p>{error}</p>
+{#if error}
+	<div class="flex items-center justify-center h-screen">
+		<p class="text-red-500">{error}</p>
 	</div>
 {:else}
-	<div class="flex items-center justify-center">
-		<div id="reader" class="w-full"></div>
+	<div class="flex items-center justify-center min-h-screen p-4">
```
VaiTon, line 60: You can't use `min-h-screen` here, as it has padding and other margins, and that makes the view scrollable and the camera view not centered vertically..

VaiTon:
Also @RISHIK92, please mention every change in the PR / title next time. You didn't only change the icon.

src/routes/qr/+page.svelte
RISHIK92: Removed it by mistake, but it shouldn’t affect the functionality. I’ll add it back.

src/routes/qr/+page.svelte
RISHIK92: I updated the container to use h-[calc(100vh-4rem)] instead of min-h-screen, so it accounts for padding and margin. while keeping the scanner centered. Let me know if any further adjustments are needed!

VaiTon: I don't like `h-[calc(100vh-4rem)]`, too complicated. Make it something like `h-full` and find a way to propagate the 100% back to the root, else just add a top margin and we're good

package-lock.json
VaiTon, line 1: Please remove this file

src/routes/qr/+page.svelte
```diff
@@ -42,19 +46,18 @@
 	});
 
 	onDestroy(() => {
-		// Stop the scanner when the component is destroyed
-		if (html5QrCode != null) {
+		if (html5QrCode) {
```
VaiTon, line 49: Please keep the check as it is

src/routes/qr/+page.svelte
```diff
@@ -42,19 +46,18 @@
...
 			html5QrCode.stop();
 		}
 	});
 </script>
 
-{#if error != null}
-	<div class="flex items-center justify-center">
-		<p>{error}</p>
+{#if error!=null}
```
VaiTon, line 55: Please run `pnpm format`

VaiTon (changes requested):

src/routes/qr/+page.svelte
```diff
@@ -42,19 +46,18 @@
...
+	<div class="flex items-center justify-center h-screen">
+		<p class="text-red-500">{error}</p>
 	</div>
 {:else}
-	<div class="flex items-center justify-center">
-		<div id="reader" class="w-full"></div>
+	<div class="flex flex-2 items-center justify-center h-[calc(100vh-4rem)]">
```
VaiTon, line 60: I don't like `h-[calc(100vh-4rem)]`, too complicated. Make it something like `h-full` and find a way to propagate the 100% back to the root, else just add a top margin and we're good

VaiTon:
@RISHIK92 with the latest commits you undoed a lot of changes. Please rebase correctly on main or open a new PR if you feel more confortable doing that.

PR #238, opened 2025-03-09
fix: visibility of barcode scan option on smaller screen devices

Prathamesh0901:
### What
The option to scan barcodes through larger screen devices (e.g. laptop) was visible, but it disappears on smaller devices like smartphones.
This PR resolves the issue by adding the button to scan barcodes in navbar.

### Screenshot

image

### Fixes bug(s)
- #224


VaiTon (approved):

VaiTon:
Thank you!

PR #242, opened 2025-03-09
corrected alignment issues

### LINKED ISSUE #230: Alignment issue in Product page in smaller screens
When we go into mobile view, you will see that the image is aligned to the right, and there's unwanted space to the the right of the details of the product

I have coloured the specific part red, so that you can have a clear view of what i am talking about

Image

The image needs to be centered in mobile view and the unwanted space at the right can be used by the details section

Bhavishya2601:
The alignment issues weren't correct on the product page on smaller screens

before (the red colour in background is just for reference)
Image

after
image

Fixes #230

### DIFF
--- a/src/lib/ui/ImageButton.svelte
+++ b/src/lib/ui/ImageButton.svelte
@@ -10,6 +10,6 @@
 
 <ImageModal bind:this={modal} />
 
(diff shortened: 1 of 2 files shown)


VaiTon (approved):
LGTM

Issue #249, opened 2025-03-10
GitHub repository link not available to mobile users in drop down menu.


swastikCommits:
### Problem
GitHub repository must also be visible to mobile users rather than just desktop users.

![Image](https://github.com/user-attachments/assets/bfb7cf3d-9c0a-401e-86af-124c35001757)

### Proposed solution
I will be adding a Button for mobile users to explore the GitHub repository.

I'm working on this.

PR #250, opened 2025-03-10
Added the GitHub icon in mobile view

swastikCommits:
### What
Added the GitHub Logo for mobile view in drop down menu

### Screenshot
Screenshot 2025-03-10 070749

### Fixes bug(s)
#249 

### Part of 
None

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -152,9 +152,21 @@
 			</button>
 		</div>
 	{/if}
-	<div class:hidden={!accordionOpen} class="mt-3 flex justify-center gap-2">
-		<a class="btn btn-outline link flex" href="/folksonomy">{$t('common.folksonomy')}</a>
-		<a class="btn btn-outline link flex" href="/settings">{$t('common.settings')}</a>
+	<div class:hidden={!accordionOpen} class="mt-3 flex flex-wrap justify-center gap-2">
+		<a class="btn btn-outline link flex items-center justify-center" href="/folksonomy">
+			{$t('common.folksonomy')}
+		</a>
+		<a class="btn btn-outline link flex items-center justify-center" href="/settings">
+			{$t('common.settings')}
+		</a>
+		<a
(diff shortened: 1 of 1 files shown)


src/routes/+layout.svelte
```diff
@@ -152,9 +152,21 @@
...
+	<div class:hidden={!accordionOpen} class="mt-3 flex flex-wrap justify-center gap-2">
+		<a class="btn btn-outline link flex items-center justify-center" href="/folksonomy">
+			{$t('common.folksonomy')}
+		</a>
+		<a class="btn btn-outline link flex items-center justify-center" href="/settings">
+			{$t('common.settings')}
+		</a>
+		<a
+			class="btn btn-outline link flex items-center justify-center"
+			href="https://github.com/openfoodfacts/openfoodfacts-explorer"
```
VaiTon, line 164: Can we factorize the URL into a const?

src/routes/+layout.svelte
```diff
@@ -152,9 +155,21 @@
...
 		</div>
 	{/if}
-	<div class:hidden={!accordionOpen} class="mt-3 flex justify-center gap-2">
-		<a class="btn btn-outline link flex" href="/folksonomy">{$t('common.folksonomy')}</a>
-		<a class="btn btn-outline link flex" href="/settings">{$t('common.settings')}</a>
+	<div class:hidden={!accordionOpen} class="mt-3 flex flex-wrap justify-center gap-2">
+		<a class="btn btn-outline link flex items-center justify-center" href="/folksonomy">
+			{$t('common.folksonomy')}
+		</a>
+		<a class="btn btn-outline link flex items-center justify-center" href="/settings">
```
VaiTon, line 162: Why do we need `flex items-center` here? It should be already handled by the `btn` class

swastikCommits: You are right actually there was no use to include that, I m fixing it.

Issue #251, opened 2025-03-10
No current percentage of slider is shown in the Influence Component in Settings page.


swastikCommits:
### What
There is no necessity to show the 100% while we actually can't see the actual value in the slider.

### Steps to reproduce the behavior:
1. Go to settings Page.
2. Checkout the "Influence" component.

### Expected behavior
Instead of 100% we cant replace it by adding a current value meter.

### Screenshots
![Image](https://github.com/user-attachments/assets/f69309af-3665-4d66-9817-10461d077ef2)

### Platform (Desktop, Mobile, Hunger Games)
 - OS: WIndows
 - Platform : Desktop

### Part of
None

I would like to work on this issue.

swastikCommits:
> What do you propose as a solution ? Showing the actual value in the slider ? Removing a label isn't going to be very additive in terms of UX.

Actually later on I decided to not remove the 100% mark, and rather added the current slider percentage in between 0% and 100%.
You can refer to the PR, I had displayed the demo of it there.

PR #252, opened 2025-03-10
Added the current percentage value of slider in Influence component

swastikCommits:
### What
Added a meter that shows the current percentage value of slider.

### Screenshot
Screenshot 2025-03-10 075308

### Fixes bug(s)
- #251 

### Part of 
None


VaiTon:
Thank you!

Issue #253, opened 2025-03-10
In Products page under Questions component the debug is overflowing.


swastikCommits:
### What
The Questions component drop down in products page in overflowing.

### Steps to reproduce the behavior:
1. Go to homepage and click any of the products.
2. Scroll down and reach the "Questions"
3. As we click the drop down the Debug code overflows in mobile view and desktop view.

### Expected behavior
Improve the responsiveness and the code should not overflow.

### Screenshots

![Image](https://github.com/user-attachments/assets/3436cd58-4aee-4ca4-9ea0-d375d8c3889e)

### Platform (Desktop, Mobile, Hunger Games)
 - OS: Windows 11
 - Platform: Mobile & Desktop

swastikCommits:
I'm currently working on this issue.

PR #257, opened 2025-03-10
Fixed the Debug component overflowing when the content is too wide

swastikCommits:
### What
1. Allow horizontal scrolling when the content is too wide
2. Enable text wrapping to prevent horizontal overflow where possible
3. Break long words that would otherwise cause overflow

### Fixes bug(s)
#253


src/lib/ui/Debug.svelte
```diff
@@ -5,5 +5,11 @@
 
 <details>
 	<summary>Debug {title ?? ''}</summary>
-	<pre class="bg-base-200 rounded-md p-2">{JSON.stringify(data, null, 2)}</pre>
+	<div class="overflow-x-auto">
+		<pre class="bg-base-200 rounded-md p-2 break-words whitespace-pre-wrap">{JSON.stringify(
```
VaiTon, line 9: Extract the JSON.stringify into a `$derived` let var

VaiTon (approved):

Issue #258, opened 2025-03-11
Knowledge Panel Action does not work


swastikCommits:
### Story
-the knowledge panel action button do not work

### Part of
#2

swastikCommits:
I have done the changes, sending a PR

teolemon:
@VaiTon we still have some cases in which it does not work

<img width="806" height="339" alt="Image" src="https://github.com/user-attachments/assets/67c4b0ad-cb12-40d4-a7f2-49cb078ad3f8" />
EG: Eco-Score, in that case: https://openfoodfacts-explorer.vercel.app/products/4013832346273

PR #259, opened 2025-03-11
Fix: knowlege panel action button not working

swastikCommits:
### What
•	Implemented a handleActionClick function that correctly processes the actions
•	Added support for common actions like "edit_product" and "report_product_to_nutripatrol"
•	Included URL detection for handling direct links in action parameters
•	Added loading states to provide visual feedback when an action is triggered
•	Implemented proper error handling if a product ID can't be found
•	Made the button display "Loading..." text while processing
•	Added special handling for buttons with HTML content
•	Ensured that links within HTML content will work properly

### Fixes bug(s)
- #258 

### Part of 
#2

### DIFF
--- a/src/lib/knowledgepanels/Element.svelte
+++ b/src/lib/knowledgepanels/Element.svelte
@@ -51,23 +51,26 @@
 		// Get the current product ID from the URL
 		const urlParts = window.location.pathname.split('/');
 		const productId = urlParts[urlParts.length - 1] || urlParts[urlParts.length - 2];
-		
+
 		// Handle different action types
 		for (const action of actions) {
-			switch(action) {
+			switch (action) {
 				case 'edit_product':
 					if (productId) {
 						window.open(`https://world.openfoodfacts.org/product/${productId}/edit`, '_blank');
 					}
 					break;
-				
+
 				case 'report_product_to_nutripatrol':
 					if (productId) {
 						// Direct to the edit page with focus on the report problem section
-						window.open(`https://world.openfoodfacts.org/product/${productId}/edit#report_problem`, '_blank');
+						window.open(
+							`https://world.openfoodfacts.org/product/${productId}/edit#report_problem`,
+							'_blank'
+						);
 					}
 					break;
-				
+
 				default:
 					// If not a known action, try to handle it as a URL if it looks like one
 					if (action.startsWith('http://') || action.startsWith('https://')) {
@@ -118,9 +121,14 @@
 			</table>
 		</div>
 	{:else if element.element_type === 'action'}
-		<button 
-			class="btn btn-primary {isLoading ? 'loading' : ''}" 
-			onclick={(event) => handleActionClick(element.action_element.actions, element.action_element.html !== '', event)}
+		<button
+			class="btn btn-primary {isLoading ? 'loading' : ''}"
+			onclick={(event) =>
+				handleActionClick(
+					element.action_element.actions,
+					element.action_element.html !== '',
+					event
+				)}
 			disabled={isLoading}
 		>
 			{#if element.action_element.html != '' && !isLoading}
--- a/src/lib/knowledgepanels/Element.svelte
+++ b/src/lib/knowledgepanels/Element.svelte
@@ -19,6 +19,8 @@
 		report_product_to_nutripatrol: 'Report Product to NutriPatrol'
 	};
 
+	let isLoading = $state(false);
+
 	function getButtonTitle(actions: string[]) {
 		for (const action of actions) {
 			if (BUTTON_ACTIONS_TITLES[action] != null) {
@@ -27,6 +29,57 @@
 		}
 
 		return actions.join(', ');
+	}
+
+	function handleActionClick(actions: string[], hasHtml: boolean, event: MouseEvent) {
+		// If the action has HTML content and the click was on a link, let the link handle it
+		if (hasHtml) {
+			const target = event.target as HTMLElement;
+			if (target.tagName === 'A' || target.closest('a')) {
+				// Let the link handle the click
+				return;
+			}
+		}
+
+		if (!actions || actions.length === 0) {
+			return;
+		}
+
+		// Show loading state
+		isLoading = true;
+
+		// Get the current product ID from the URL
+		const urlParts = window.location.pathname.split('/');
+		const productId = urlParts[urlParts.length - 1] || urlParts[urlParts.length - 2];
+		
+		// Handle different action types
+		for (const action of actions) {
+			switch(action) {
+				case 'edit_product':
+					if (productId) {
+						window.open(`https://world.openfoodfacts.org/product/${productId}/edit`, '_blank');
+					}
+					break;
+				
+				case 'report_product_to_nutripatrol':
+					if (productId) {
+						// Direct to the edit page with focus on the report problem section
+						window.open(`https://world.openfoodfacts.org/product/${productId}/edit#report_problem`, '_blank');
+					}
+					break;
+				
+				default:
+					// If not a known action, try to handle it as a URL if it looks like one
+					if (action.startsWith('http://') || action.startsWith('https://')) {
+						window.open(action, '_blank');
+					}
+			}
+		}
+
+		// Reset loading state after a short delay
+		setTimeout(() => {
+			isLoading = false;
+		}, 1000);
 	}
 </script>
 
@@ -65,12 +118,18 @@
 			</table>
 		</div>
 	{:else if element.element_type === 'action'}
-		<button class="btn btn-primary">
-			{#if element.action_element.html != ''}
+		<button 
+			class="btn btn-primary {isLoading ? 'loading' : ''}" 
+			onclick={(event) => handleActionClick(element.action_element.actions, element.action_element.html !== '', event)}
+			disabled={isLoading}
+		>
+			{#if element.action_element.html != '' && !isLoading}
 				<!-- eslint-disable-next-line svelte/no-at-html-tags -->
 				{@html element.action_element.html}
-			{:else}
+			{:else if !isLoading}
 				{getButtonTitle(element.action_element.actions)}
+			{:else}
+				Loading...
 			{/if}
 		</button>
 	{:else if element.element_type === 'map'}


VaiTon:
I think the best way forward here is to extract the specific KP into its own component, instead of using the Element generic one

VaiTon:
@swastikCommits why not an `Action.svelte` components that only handles Action KPs?

swastikCommits:
> @swastikCommits why not an `Action.svelte` components that only handles Action KPs?

Yes I did the same but its KnowledgePanelElement.svelte.
@VaiTon if you say I can rename it to Action.svelte. Please review this commit

VaiTon:
> Yes I did the same but its KnowledgePanelElement.svelte.

@swastikCommits we already have an `Element.svelte` inside `lib/knowledgepanels`. Adding another one with the same name is not good.

VaiTon:
Okay I saw the diff... What I was telling you is that you should extract the logic you added into a component.

swastikCommits:
Ok I got it now. So, @VaiTon  shall I delete the KnowledgePanelElement.svelte and instead make another component called Action.svelte which would take care of the function handleActionClick and the loading state?

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,118 @@
+<script lang="ts">
+	import type {
+		KnowledgeActionElement,
+		KnowledgePanel,
+		KnowledgePanelElement as KnowledgePanelElementType
+	} from '$lib/api';
+	import Panel from './Panel.svelte';
+
+	type Props = {
+		allPanels?: Record<string, KnowledgePanel>;
```
VaiTon, line 10: Make this non-nullable

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,118 @@
...
+			<!-- eslint-disable-next-line svelte/no-at-html-tags -->
+			{@html element.action_element.html}
+		{:else if !isLoading}
+			{getButtonTitle(element.action_element.actions)}
+		{:else}
+			Loading...
+		{/if}
+	</button>
+{:else if element.element_type === 'panel'}
+	{#if allPanels}
```
VaiTon, line 105: This cannot be null (initialized by default)

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,118 @@
...
+		{#if 'panel_element' in element}
+			{@const id = element.panel_element.panel_id}
+			{@const panel = allPanels[id]}
+			{#if panel}
```
VaiTon, line 109: Invert this if and use the `!== null` syntax

VaiTon (changes requested):

swastikCommits:
@VaiTon please verify the changes I resolved

src/lib/knowledgepanels/Action.svelte
swastikCommits: `import Panel from './Panel.svelte';

	type Props = {
		allPanels: Record<string, KnowledgePanel>;
		element: KnowledgeActionElement | KnowledgePanelElementType;
	};
	let { allPanels, element }: Props = $props();
`

src/lib/knowledgepanels/Action.svelte
swastikCommits: omitted `{#if allPanels}`

src/lib/knowledgepanels/Action.svelte
swastikCommits: `
		{#if panel !== null}
			<Panel {panel} {allPanels} {id} />
`

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,114 @@
...
+		if (!actions || actions.length === 0) {
+			return;
+		}
+
+		// Show loading state
+		isLoading = true;
+
+		// Get the current product ID from the URL
+		const urlParts = window.location.pathname.split('/');
+		const productId = urlParts[urlParts.length - 1] || urlParts[urlParts.length - 2];
```
VaiTon, line 52: This is not a safe method to get the product code. What if the URL changes? Please create a property on the component and pass it from the `+page.svelte`

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,114 @@
...
+			isLoading = false;
+		}, 1000);
+	}
+</script>
+
+{#if element.element_type === 'action'}
+	<button
+		class="btn btn-primary {isLoading ? 'loading' : ''}"
+		onclick={(event) =>
+			handleActionClick(element.action_element.actions, element.action_element.html !== '', event)}
```
VaiTon, line 92: `element` is a property, so we can directly use it inside the function instead of passing it in this call.

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,114 @@
...
+
+	function handleActionClick(actions: string[], hasHtml: boolean, event: MouseEvent) {
+		// If the action has HTML content and the click was on a link, let the link handle it
+		if (hasHtml) {
+			const target = event.target as HTMLElement;
+			if (target.tagName === 'A' || target.closest('a')) {
+				// Let the link handle the click
+				return;
+			}
+		}
```
VaiTon, line 41: In what cases do we need this? Can we document it?

VaiTon (changes requested):

src/lib/knowledgepanels/Action.svelte
swastikCommits: updated the Action.svelte component to accept a productCode prop instead of extracting it from the URL, and updated the component tree to pass this prop down from the product page

src/lib/knowledgepanels/Action.svelte
swastikCommits: You're absolutely right. Since `element `is already a property of the component, we can simplify the code by accessing it directly inside the `handleActionClick `function rather than passing parts of it as arguments

src/lib/knowledgepanels/Action.svelte
swastikCommits: Some knowledge panels contain rich HTML content with clickable links inside the action button.
so, whenever a user clicks directly on a link within the HTML content, we want the link to function normally (navigate to its destination) rather than triggering the button's action.

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,126 @@
...
+		// Show loading state
+		isLoading = true;
+
+		// Use the productCode prop instead of extracting from URL
+		// Handle different action types
+		for (const action of element.action_element.actions) {
+			switch (action) {
+				case 'edit_product':
+					if (productCode) {
+						window.open(`https://world.openfoodfacts.org/product/${productCode}/edit`, '_blank');
```
VaiTon, line 72: Extract in a const

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,126 @@
...
+					}
+					break;
+
+				case 'report_product_to_nutripatrol':
+					if (productCode) {
+						// Direct to the edit page with focus on the report problem section
+						window.open(
+							`https://world.openfoodfacts.org/product/${productCode}/edit#report_problem`,
```
VaiTon, line 80: Extract in a const

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,126 @@
...
+			return;
+		}
+
+		// Show loading state
+		isLoading = true;
+
+		// Use the productCode prop instead of extracting from URL
+		// Handle different action types
+		for (const action of element.action_element.actions) {
+			switch (action) {
```
VaiTon, line 69: As we have a lot of conditions here, use an if-else statement and collapse the nesting

VaiTon (changes requested):

src/lib/knowledgepanels/Element.svelte
```diff
@@ -7,36 +7,21 @@
...
 
 <div class="my-1">
 	{#if element.element_type === 'panel_group'}
-		<PanelGroup {element} {allPanels} />
-	{:else if element.element_type === 'panel'}
-		{@const id = element.panel_element.panel_id}
-		{@const panel = allPanels[id]}
-		<Panel {panel} {allPanels} {id} />
+		<PanelGroup {element} {allPanels} {productCode} />
+	{:else if element.element_type === 'panel' || element.element_type === 'action'}
```
VaiTon, line 23: I'm not sure to understand why we're handling the two cases here. Can't we handle the single `action` case with the `Action.svelte` component and leave the `panel` here?

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		// Show loading state
+		isLoading = true;
+
+		// Get the first action from the list (most cases will only have one action)
+		const action = element.action_element.actions[0];
+
+		// Handle known action types if product code is available
+		if (productCode) {
```
VaiTon, line 75:
```suggestion
		if (productCode != null) {
```

VaiTon (changes requested):
Sorry for the delay! Just a few things and we're done!

src/lib/knowledgepanels/Element.svelte
swastikCommits: In Element.svelte, we're routing both 'panel' and 'action' elements to the Action.svelte component
But in Action.svelte, we already handle each case separately

It would make more sense to separate these concerns:
1. Keep Action.svelte for handling actions only
2. Handle the panel case directly in Element.svelte

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,116 @@
+<script lang="ts">
+	import type { KnowledgeActionElement } from '$lib/api';
+
+	// URL constants
+	const PRODUCT_EDIT_URL = 'https://world.openfoodfacts.org/product/';
+	const PRODUCT_REPORT_URL = 'https://world.openfoodfacts.org/product/';
+
+	type Props = {
+		element: KnowledgeActionElement;
+		allPanels: Record<string, any>; // Keep for compatibility
```
VaiTon, line 10: `any` is a no-go. What do you mean "for compatibility"?

swastikCommits: this prop is kept for backward compatibility with existing code but isn't properly typed.
The `allPanels` is passed to `Action.svelte `from `Element.svelte`, but it isn't actually used in the Action component at all. It's defined in the props but not destructured for use.

swastikCommits: `allPanels: Record<string, KnowledgePanel>;`

VaiTon: If it's not used you should remove that

swastikCommits: Its used actually, I had made the changes by removing `any`

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,116 @@
...
+		isLoading = true;
+
+		// Get the first action from the list (most cases will only have one action)
+		const action = element.action_element.actions[0];
+
+		// Handle known action types if product code is available
+		if (productCode != null) {
+			// Edit product action
+			if (action === 'edit_product') {
+				window.open(`${PRODUCT_EDIT_URL}${productCode}/edit`, '_blank');
```
VaiTon, line 69: We should stay in our frontend to edit the product :)). Use the `goto` func

src/lib/knowledgepanels/Action.svelte
```diff
@@ -0,0 +1,116 @@
...
+			}
+			// Report to NutriPatrol action
+			else if (action === 'report_product_to_nutripatrol') {
+				window.open(`${PRODUCT_REPORT_URL}${productCode}/edit#report_problem`, '_blank');
```
VaiTon, line 73: We can do this for now, but please add a TODO explaining that we should eventually just link to the right route for this.

VaiTon (changes requested):

src/lib/knowledgepanels/Action.svelte
swastikCommits: Sure, Done 👍

Issue #265, opened 2025-03-12
X logo is not correct in the footer, old twitter's logo is used


Prathamesh0901:
the footer created contains the twitter's logo not X though the platform's name is X now
![Image](https://github.com/user-attachments/assets/dd83cff5-322c-4ccd-88b3-70715a173fae)

Prathamesh0901:
open to work

swastikCommits:
I would like to work on it.

VaiTon:
I mean, yes... but it's really not something I would spend time on

PR #267, opened 2025-03-12
Replaced the twitter logo with X logo

swastikCommits:
### What
Replaced the Twitter older logo with the newer X logo.

The MDI class we use to extract the logos doesn't have the new X-logo, so I created a X-logo.svg in 
``static/images/icons/x-logo.svg``

### Screenshot
Screenshot 2025-03-12 143747

### Fixes bug(s)
#265 

### Part of 
None

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
 	"devDependencies": {
 		"@eslint/eslintrc": "^3.3.0",
 		"@eslint/js": "^9.22.0",
+		"@iconify-json/logos": "^1.2.4",
 		"@iconify-json/mdi": "^1.2.3",
 		"@iconify/tailwind": "^1.2.0",
 		"@iconify/tailwind4": "^1.0.6",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -14,6 +14,9 @@ importers:
       '@eslint/js':
         specifier: ^9.22.0
         version: 9.22.0
+      '@iconify-json/logos':
+        specifier: ^1.2.4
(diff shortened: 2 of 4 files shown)


VaiTon:
I don't really like keeping svgs in-tree. If the alternative is to keep the old logo I'm ok with it. Otherwise let's find an icon set that has the new icon.

swastikCommits:
> I don't really like keeping svgs in-tree. If the alternative is to keep the old logo I'm ok with it. Otherwise let's find an icon set that has the new icon.

imo keeping the new logo instead of old one would be better.
If we don't want to use svgs then let me find some logo set again instead of MDI

swastikCommits:
@VaiTon found **FontAwesome** for the X-logo.
What do you want me to do?
1. Install dependency for FontAwesome and import it to `Footer.svelte`
2. Using CDN in `app.html`
3. We can directly add the svg in the that div of `Footer.svelte`

swastikCommits:
Removed the .svg from static folder and directly implemented it in the attribute itself.

**This will solve the issue temporarily, until we find some logo sets containing extensive variety of updated logos**

VaiTon:
> Removed the .svg from static folder and directly implemented it in the attribute itself.

You should find an icon from an iconset of the framework we use (https://iconify.design/) and use that. Embedding the svg is no better than having it as a static file.

swastikCommits:
@VaiTon have a look added the dependency iconify-json/logos and imported it for x-logo

VaiTon:
Thanks! :tada:

Issue #113, opened 2024-09-11
Add support for Open Beauty Facts, Open Pet Food Facts, Open Products Facts which now provide the latest APIs (incl. Knowledge Panels)


teolemon:

KarruHarin:
I want to first try to integrate OpenBeautyFacts Api and then move on to others .I would like to be assigned .

teolemon:
![Image](https://github.com/user-attachments/assets/bd4133f8-d504-43de-b544-164b35b988b2)
Trying to enter a Cosmetic barcode does not work: https://world.openbeautyfacts.org/product/3245678674840/bain-de-bouche-fraicheur-intense-auchan
https://openfoodfacts-explorer.vercel.app/products/3245678674840

KarruHarin:
![Image](https://github.com/user-attachments/assets/2b26ff41-26cf-48f1-ad38-01e10d442c2a)

@VaiTon some beauty products are listed & some are not ,does that mean there are some beauty products in the OFF data?
and can I get a little guidence on how to get started ? I am looking forward to add support for OBF :)

VaiTon:
> does that mean there are some beauty products in the OFF data

Unfortunately, yes. AFAIK we're only querying the OFF endpoints, so no OBF data should be displayed.

> Can I get a little guidence on how to get started ? I am looking forward to add support for OBF

@teolemon do we have a way of querying if a certain barcode is of a certain type? (other than trying to request it and having an error in the response?)

KarruHarin:
@teolemon I want to add a search option for OBF , like a small drop down menu to either select OFF or OBF , then they can search in that respectively . Is that okay ? or do you have any suggestions on how to implement it in search .

Bhavishya2601:
@VaiTon @teolemon what we can do is first find it in open food facts and if it is not found then we can search it in open beauty facts (right now I tried with beauty API only), I think this method is better then giving a dropdown to the user to select OFF or OBF.
I tried implementing it, you can see below

https://github.com/user-attachments/assets/0df49b84-acc0-4042-a87b-d2e9c71e1104

It is not able to show the data because it is configured according to OFF

BunnyKarru:
@Bhavishya2601 but consider this when you search for beauty products you want to get results from OBF , but there are some beauty products in OFF too.
[...]

PR #269, opened 2025-03-12
OBF Integration to Product Page

KarruHarin:
### What
Added OBF integration into the product page so no error appears even when we try for OBF Products
Product Page now supports OBF
### Screenshot
obf2
obf1
These are OBF products and not OFF Products

### Part of 
#113 

###Next Steps
I have added OBF to product page , working on adding it to edit and search too.

Looking forward for you feedback or any modifications !

Thank you :)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -4,3 +4,73 @@
 GitHub Actions:
   - changed-files:
       - any-glob-to-any-file: '.github/**/*'
+🏷️ Folksonomy Project:
+  - changed-files:
+      - any-glob-to-any-file: 'src/routes/folksonomy/+page.svelte'
+📖 Knowledge panels:
+  - changed-files:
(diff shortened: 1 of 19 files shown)


teolemon:
We should customize behaviours, eg hide Nutri-Score, Nova, and Eco-Score.

BunnyKarru:
> We should customize behaviours, eg hide Nutri-Score, Nova, and Eco-Score.

Do you have any other customisations too in mind ! Gonna apply them and raise a updated PR.

VaiTon:
I'm sorry but I don't think this PR has a very good resulting architecture. We should first discuss and think thoroughly about how to add the other databases to the app.

The fact that this design is not scalable is the fact that you already had to put an {#if} block to just display a different text / base url. Imagine how many if-elses we will have to add to render each case.

Issue #262, opened 2025-03-11
Perform a universal search (OxF) from the barcode scanner.


lakshaydewan:
Story
Perform a universal search (OxF) from the barcode scanner.
also Add a way to add a product if it does not exist on a universal search (OxF)

Part of
#2

github-throwaway:
@teolemon what did you mean with "Perform a universal search (OxF) from the barcode scanner" ? Could you clarify.

github-throwaway:
@teolemon what did you mean with "Perform a universal search (OxF) from the barcode scanner" ? Could you clarify.

github-throwaway:
> If you add &product_type=all to your query, you will get redirected


https://openfoodfacts.slack.com/archives/C08FMJWP6J0/p1744106491016999?thread_ts=1743728815.538489&channel=C08FMJWP6J0&message_ts=1744106491.016999

PR #271, opened 2025-03-12
Added universal search OxF from barcode scanner.

lakshaydewan:
What
Implemented OxF search from the qr route and added not found implementation

Screenshot
Screenshot 2025-03-12 at 4 44 23 PM

Fixes bug(s)
#262 

Part of 
#2

### DIFF
--- a/src/routes/qr/+page.svelte
+++ b/src/routes/qr/+page.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import { onDestroy, onMount } from 'svelte';
 	import { Html5Qrcode, Html5QrcodeSupportedFormats } from 'html5-qrcode';
+	import { PRODUCT_URL } from '$lib/const';
 
 	let error: string | null = $state(null);
 	let productNotFound: boolean = $state(false);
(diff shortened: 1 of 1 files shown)


lakshaydewan:
@VaiTon waiting for review. Plus should i start implementing an add-product route if this is fine?

src/routes/qr/+page.svelte
```diff
@@ -3,9 +3,30 @@
...
 
 	let error: string | null = $state(null);
+	let productNotFound: boolean = $state(false);
+	let scannedBarcode: string | null = $state(null);
 
 	let html5QrCode: Html5Qrcode | null = null;
 
+	async function searchProduct(barcode: string) {
+		try {
+			const response = await fetch(`https://world.openfoodfacts.org/api/v2/product/${barcode}.json`);
```
VaiTon, line 13: Please don't use hardcoded urls. See how API requests are made in the `+page.ts` files

VaiTon (changes requested):

src/routes/qr/+page.svelte
lakshaydewan: Will change this plus what about the add product page?

lakshaydewan:
@VaiTon fixed the URL plz review.

VaiTon:
Architectural-wise, this is not how I wanted it to be.

The `.svelte` file should not make any request ideally. We should handle this case in the product page, not in the QR one.

lakshaydewan:
@VaiTon well in that case the issue should be removed from #2.

VaiTon:
The issue persists. The problem is that the software component which should be responsible of this is not the code scanner. 

It should be done in a +page.ts, and then handled visually in the +page.svelte file of the product display.

lakshaydewan:
Got it!

Issue #268, opened 2025-03-12
Header is unsymmetrical and gets scrolled with the page


swastikCommits:
### Problem

1. The padding on the header mismatches with the padding on footer so it looks unsymmetrical with respect to the footer.
2. Header gets scrolled as we scroll the page, `position:fixed` is missing.

### Proposed solution
Increasing the padding on left side and right side of the header.

![Image](https://github.com/user-attachments/assets/2dd70839-1c31-409b-a5ea-716f853889a1)

Referring from https://world.openfoodfacts.org/ 
![Image](https://github.com/user-attachments/assets/17d114bd-3ae0-46fc-bf2c-57a7551c0f53)

swastikCommits:
I'm currently working on this.

VaiTon:
I didn't understand this issue, and I'm not quite sure I agree with it

swastikCommits:
> I didn't understand this issue, and I'm not quite sure I agree with it

Before:![Image](https://github.com/user-attachments/assets/a8bc37c7-b5a1-451f-bd4d-fed4afa7c5b1)




After:![Image](https://github.com/user-attachments/assets/2c8bc102-7452-4e2e-bf99-3518c9f76e40)



@VaiTon Now you can see the homepage looks symmetrical.
I found another issue with header that it doesn't have the class `position:fixed`  so, shall I fix it in the same PR?

PR #275, opened 2025-03-13
feat: added padding to header for symmetrical UI

swastikCommits:
### What
Added left and right padding to header, making it symmetrical to the footer.
Desktop Header (md+): stays above content with shadow.
Mobile Header (<md): adjusted alignment.

### Screenshot
Screenshot 2025-03-13 093302

### Fixes bug(s)
#268 

### Part of 
None

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -49,9 +49,7 @@
 	<meta name="viewport" content="width=device-width, initial-scale=1" />
 </svelte:head>
 
-<div
-	class="bg-base-100 navbar top-0 right-0 left-0 z-50 hidden px-10 md:flex md:px-20 lg:px-40"
->
+<div class="bg-base-100 navbar hidden px-10 md:flex md:px-20 lg:px-40">
 	<div class="navbar-start">
 		<a href="/"> <Logo /> </a>
 	</div>
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -50,7 +50,7 @@
 </svelte:head>
 
 <div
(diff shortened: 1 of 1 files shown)


src/routes/+layout.svelte
```diff
@@ -49,7 +49,9 @@
 	<meta name="viewport" content="width=device-width, initial-scale=1" />
 </svelte:head>
 
-<div class="navbar hidden md:flex">
+<div
+	class="bg-base-100 navbar fixed top-0 right-0 left-0 z-50 hidden px-10 shadow-md md:flex md:px-20 lg:px-40"
```
VaiTon, line 53: Can we remove the `fixed` class? I don't like it very much :/

VaiTon (changes requested):

src/routes/+layout.svelte
swastikCommits: Sir, but it would create a better User Experience.

No worries removing it :)

VaiTon (changes requested):

src/routes/+layout.svelte
```diff
@@ -49,7 +49,9 @@
 	<meta name="viewport" content="width=device-width, initial-scale=1" />
 </svelte:head>
 
-<div class="navbar hidden md:flex">
+<div
+	class="bg-base-100 navbar top-0 right-0 left-0 z-50 hidden px-10 md:flex md:px-20 lg:px-40"
```
VaiTon, line 53: Do we need top-0, right-0, left-0 and z-50 now?

swastikCommits: my bad, as we removed the `fixed` class then there would be no need of it anymore

VaiTon (approved):

VaiTon:
Thank you! :rocket:

Issue #260, opened 2025-03-11
Facets tasks


swastikCommits:
### Story

- [x] Beautify the Folksonomy Engine key display
- [ ] Implement the Knowledge Panel Facets for facets, preferably using a webcomponent - https://github.com/openfoodfacts/facets-knowledge-panels
- [ ] Add support for https://github.com/openfoodfacts/openfoodfacts-query for Facets

### Part of
#2

swastikCommits:
I'm currently working to fix all the 3 sub tasks.
@VaiTon shall I combine all the sub tasks and make a single PR or make separate 3 PRs?

VaiTon:
> preferably using a webcomponent 

Mmmm, I'm not really sure about this. Let's try, but I reserve the right to reject the PR

PR #288, opened 2025-03-14
feat: beautification of Folksonomy engine key (/folksonomy)

swastikCommits:
### What

- Upgraded Visuals: Added a sleek header with subtitle, switched to a card-based layout, and included progress bars, colored badges, hover animations, and a responsive grid with smooth transitions.
- Better Structure: Keys are auto-grouped by prefix, collapsible for less clutter, and feature usage badges with a clear visual hierarchy.
- New Features: Introduced a search box with result counter, empty state message, and usage progress bars.
- Tech Enhancements: Built with Svelte’s reactive variables, optimized rendering, and accessible, responsive design.

### Screenshot

https://github.com/user-attachments/assets/bd0f7e6b-0319-4012-b936-a82f33fb0f33

### Fixes bug(s)
#260 #315 

### Part of 
#2

### DIFF
--- a/src/routes/folksonomy/+page.svelte
+++ b/src/routes/folksonomy/+page.svelte
@@ -25,8 +25,8 @@
 
 	// Grouping functionality
 	function getKeyPrefix(key: string): string {
-		// Extract the first part of the key before a colon or underscore
-		const match = key.match(/^([^:_]+)/);
+		// Extract the first part of the key before a colon
+		const match = key.match(/^([^:]+)/);
 		return match != null ? match[0] : 'other';
 	}
 
@@ -101,11 +101,6 @@
 		if (percent > 25) return 'badge-accent';
 		return 'badge-ghost';
 	}
-
-	function handleSearch() {
-		// This function can be used for additional search functionality if needed
-		console.log('Searching for:', searchQuery);
-	}
 </script>
 
 <div class="folksonomy-container flex flex-col gap-6 p-4" transition:fade={{ duration: 500 }}>
@@ -131,11 +126,7 @@
 							bind:value={searchQuery}
 							transition:fade={{ duration: 200 }}
 						/>
-						<button
-							class="btn btn-square"
-							aria-label="Search folksonomy keys"
-							onclick={handleSearch}
-						>
+						<button class="btn btn-square" aria-label="Search folksonomy keys">
 							<span class="icon-[mdi--magnify] h-6 w-6"></span>
 						</button>
 					</div>
@@ -175,7 +166,7 @@
 								class:badge-secondary={usagePercent > 50 && usagePercent <= 75}
 								class:badge-accent={usagePercent > 25 && usagePercent <= 50}
 								class:badge-ghost={usagePercent <= 25}
-								title="Number of products: {key.count}"
+								title="Used in {key.count} products ({usagePercent}% relative usage)"
 							>
 								{key.count}
 							</div>
@@ -183,7 +174,7 @@
 						{#if key.values != null && key.values > 0}
 							<div
 								class="badge badge-sm {getColorClass(getValuesPercent(key.values))}"
-								title="Number of unique values: {key.values}"
+								title="Has {key.values} different possible values"
 							>
 								{key.values}
 							</div>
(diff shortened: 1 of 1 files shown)


VaiTon:
Thank you for the PR! 

Just a few things:
- Could you please remove any inline svg? Instead you can use Iconify to add icons.
- Make sure to use the `if (a != null)` syntax instead of `if (a)`
- Try to split the rendering in `snippets`

src/routes/folksonomy/+page.svelte
```diff
@@ -1,21 +1,277 @@
...
+	);
+
+	const maxValues = $derived(
+		data.keys != null && data.keys.length > 0
+			? Math.max(...data.keys.map((k) => k?.values || 0), 1)
+			: 1
+	);
+
+	function getUsagePercent(count: number): number {
+		return Math.ceil(((count || 0) / maxCount) * 100);
```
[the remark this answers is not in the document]
swastikCommits: To address this, I should:
1. Update the function parameter type to accurately reflect that it can be null or undefined
2. Keep the` || 0` fallback since it's handling potential null/undefined values

It would be better that the proper type should be `number | null | undefined`

src/routes/folksonomy/+page.svelte
```diff
@@ -1,21 +1,277 @@
...
+				</div>
+			{/each}
+		</div>
+	{/if}
+</div>
+
+<style>
+	.folksonomy-container {
+		animation: fadeIn 0.5s ease-in-out;
+	}
```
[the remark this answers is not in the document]
swastikCommits: Yeah sure! 
The `fade` transition is already being used in the code, so we can apply it to the main container element instead of using the custom CSS animation.

src/routes/folksonomy/+page.svelte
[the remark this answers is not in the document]
VaiTon: > Update the function parameter type to accurately reflect that it can be null or undefined

Please analyze when it can be null / undefined and treat the problem at its root. The `getUsagePercent` function should never receive a null.

> Keep the || 0 fallback since it's handling potential null/undefined values

Nah, I prefer it to crash if we get null values. At least we would instantly where the problem is xD

swastikCommits:
@VaiTon have a look is it fine now?

src/routes/folksonomy/+page.svelte
```diff
@@ -1,21 +1,267 @@
...
+	let filteredTags = $derived.by(() => {
+		return searchQuery != null && searchQuery !== ''
+			? sortedTags.filter((key) => key.k.toLowerCase().includes(searchQuery.toLowerCase()))
+			: sortedTags;
+	});
+
+	// Grouping functionality
+	function getKeyPrefix(key: string): string {
+		// Extract the first part of the key before a colon or underscore
+		const match = key.match(/^([^:_]+)/);
```
VaiTon, line 29: I think we should only split on `:`, and not also on `_`

src/routes/folksonomy/+page.svelte
```diff
@@ -1,21 +1,267 @@
...
+		if (percent > 75) return 'badge-primary';
+		if (percent > 50) return 'badge-secondary';
+		if (percent > 25) return 'badge-accent';
+		return 'badge-ghost';
+	}
+
+	function handleSearch() {
+		// This function can be used for additional search functionality if needed
+		console.log('Searching for:', searchQuery);
+	}
```
VaiTon, line 108: Was this made by you :eyes: ?

VaiTon:
immagine

As a user, I would not understand what do the progress bar and the badges mean. Can we remove them / explain them in the UI?

src/routes/folksonomy/+page.svelte
swastikCommits: Ah sorry I was stuck so took a little bit of help to debug :(

src/routes/folksonomy/+page.svelte
swastikCommits: Done 👍

swastikCommits:
@VaiTon 

> immagine
> 
> As a user, I would not understand what do the progress bar and the badges mean. Can we remove them / explain them in the UI?

I have added explanation in the top of the page and on the hover too

VaiTon (approved):

VaiTon:
Thank you! :tada:

Issue #170, opened 2025-03-01
Implement packaging editing


VaiTon:

KrownWealth:
I would love to work on this, can you assign it to me?

VaiTon:
Hey @KrownWealth, any news on this one?

swastikCommits:
@VaiTon this has been solved in #291 
So you can close this issue now.

teolemon:
Reopening this one, because what we'd like is:
- structured packaging editing
- if possible using openfoodfacts-nodejs
- if possible as a webcomponent in https://github.com/openfoodfacts/openfoodfacts-webcomponents
- An implementation exists on the website
- An implementation exists in Hunger Games

Issue #168, opened 2025-03-01
Implement basic details editing


VaiTon:

Sharma-Ji-21:
@VaiTon pls assign this issue to me i want to work on it

VaiTon:
@Sharma-Ji-21 done!

Bhavishya2601:
@Sharma-Ji-21 Are you active on this issue?

VaiTon:
@teolemon what's missing of this one?

github-throwaway:
Closing. Please reopen with more info if required.

Issue #114, opened 2024-09-11
Implement nutrition editing


teolemon:

teolemon:
We now have a system to auto-extract nutrition

YourBroCode:
Hey @teolemon can i work on it ?

VaiTon:
@YourBroCode sure!

YourBroCode:
Dear mentor, @VaiTon could you please help me understand the issue better.
do we need to implement the editing of nutrition sample of the product or the nutritional facts:
like this one : -
<img width="1112" alt="Image" src="https://github.com/user-attachments/assets/f96948f3-cc87-4316-91ce-2535edaa5289" />

or this one :

<img width="991" alt="Image" src="https://github.com/user-attachments/assets/6f077e78-8421-4061-99ff-f9203fd32d15" />

VaiTon:
Hey @YourBroCode, we need to implement a way to edit the second image you posted, the nutrient table. You can see a similar table if you try to edit a product on openfoodfacts.org.

teolemon:
@VaiTon @SyedImtiyaz-1 @alexgarel @Kout95 we have a webcomponent in the https://github.com/openfoodfacts/openfoodfacts-webcomponents repository. 
We should really try, given the complexity to mutualize efforts.

github-throwaway:
Closing. Please reopen with more info if required.

PR #291, opened 2025-03-15
feat: implementation of product basic editing, packaging editing and nutritional information editing

swastikCommits:
### What
1. Added new fields to the edit form:
- EMB code (Producer code)
- Packaging information
- Manufacturing places
- Serving size

2. Added a comprehensive nutritional information editing section that allows users to edit:

- Energy values (kJ and kcal)
- Fat and saturated fat
- Carbohydrates and sugars
- Proteins
- Salt and sodium

3. Updated the Product type definition to include the new fields

4. Updated the API submission method to include these new fields when saving the product

### Screenshot

https://github.com/user-attachments/assets/af47b591-a98f-4679-957b-e5fc7b140258

### Fixes bug(s)
#168 
#170 
#114 

### Part of 
#125

### DIFF
--- a/src/routes/products/[barcode]/edit/+page.svelte
+++ b/src/routes/products/[barcode]/edit/+page.svelte
@@ -145,6 +145,25 @@
 	</div>
 
 	<div class="form-control mb-4">
+		<label for="">EMB Code (Producer code)</label>
+		<input type="text" class="input input-bordered w-full" bind:value={$productStore.emb_codes} />
+	</div>
+
+	<div class="form-control mb-4">
+		<label for="">Packaging</label>
+		<input type="text" class="input input-bordered w-full" bind:value={$productStore.packaging} />
+	</div>
+
+	<div class="form-control mb-4">
+		<label for="">Manufacturing places</label>
+		<input
+			type="text"
+			class="input input-bordered w-full"
+			bind:value={$productStore.manufacturing_places}
+		/>
+	</div>
+
+	<div class="form-control mb-4">
 		<label for="">Categories: </label>
 		<TagsString bind:tagsString={$productStore.categories} autocomplete={categoryNames} />
 	</div>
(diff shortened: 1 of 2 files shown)


swastikCommits:
@VaiTon Please review it, made a major feature update which would close 3 issues in this PR :)

src/routes/products/[barcode]/edit/+page.svelte
```diff
@@ -206,6 +225,97 @@
...
+				<label for="">Sodium (g)</label>
+				<input
+					type="number"
+					class="input input-bordered w-full"
+					bind:value={$productStore.nutriments.sodium_100g}
+				/>
+			</div>
+		</div>
+	{:else}
+		<div class="alert alert-warning">No nutritional information available</div>
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/+page.svelte:315]: We should still be able to edit them by inserting them. The only case in which we should not display this table is when the "No nutrition information on the product" checkbox is checked (see the main website).

swastikCommits:
@VaiTon 


Implemented the checkbox for nutritional info: 
- Unchecked: shows nutritional info form.
- Checked: displays a message hiding the info.
- Supports adding values to an initially empty nutriments object.

src/routes/products/[barcode]/edit/+page.svelte
```diff
@@ -4,14 +4,26 @@
...
 
 	import type { PageData } from './$types';
 	import TagsString from './TagsString.svelte';
 	import { PRODUCT_IMAGE_URL } from '$lib/const';
 
+	// Extend Product type to include no_nutrition_data property
+	type ExtendedProduct = Product & {
+		no_nutrition_data?: boolean;
+	};
+
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/+page.svelte:26]: Why should this be a different type from Product? Can't we add the required property to Product?

src/routes/products/[barcode]/edit/+page.svelte
```diff
@@ -236,83 +273,135 @@
...
 				<label for="">Fat (g)</label>
 				<input
 					type="number"
 					class="input input-bordered w-full"
-					bind:value={$productStore.nutriments.fat_100g}
+					value={$productStore.nutriments?.fat_100g ?? ''}
+					oninput={(e) =>
+						updateNutriment(
+							'fat_100g',
+							e.currentTarget.value ? Number(e.currentTarget.value) : null
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/+page.svelte:320]: Factorize this call into a method that gets the event and the key.

VaiTon (changes requested):

src/routes/products/[barcode]/edit/+page.svelte
Comment (swastikCommits) [reply src/routes/products/[barcode]/edit/+page.svelte:26]: The `ExtendedProduct` type is being created because the original Product type likely comes from an imported library `$lib/api` 
Creating an extension type allows adding the `no_nutrition_data` property specifically for this component's use without modifying the original type definition.

But as you requested for the changes, I am including the type `no_nutrition_data` in `product.ts`

src/routes/products/[barcode]/edit/+page.svelte
Comment (swastikCommits) [reply src/routes/products/[barcode]/edit/+page.svelte:320]: I've created a `handleNutrimentInput` function that accepts an event and a nutriment key, making the code more concise and maintainable. The function properly extracts the value from the input field and passes it to the existing `updateNutriment` function.

src/routes/products/[barcode]/edit/+page.svelte
Comment (VaiTon) [reply src/routes/products/[barcode]/edit/+page.svelte]: I mean, it is something that is part of the product, not something only this component should access. The types are everything but complete, so feel free to modify them in the futures =)

src/routes/products/[barcode]/edit/+page.svelte
```diff
@@ -34,7 +41,7 @@
 	let storeNames = $derived(getNames(data.stores));
 	let originNames = $derived(getNames(data.origins));
 	let countriesNames = $derived(getNames(data.countries));
-	let productStore = $derived(writable(data.state.product));
+	let productStore = $derived(writable<Product>(data.state.product));
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/+page.svelte:44]: Do we need to specify it?

src/routes/products/[barcode]/edit/+page.svelte
```diff
@@ -206,6 +262,113 @@
...
+				<input
+					type="number"
+					class="input input-bordered w-full"
+					value={$productStore.nutriments?.sodium_100g ?? ''}
+					oninput={(e) => handleNutrimentInput(e, 'sodium_100g')}
+				/>
+			</div>
+		</div>
+	{:else}
+		<div class="alert alert-info">Nutrition information will not be displayed for this product</div>
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/+page.svelte:368]: Use the {$t(...)} for i18n. It should be something like "Nutritional infos are not specified for this product"

VaiTon (commented):
Just two nitpicks and we're good to go!

src/routes/products/[barcode]/edit/+page.svelte
Comment (swastikCommits) [reply src/routes/products/[barcode]/edit/+page.svelte:44]: - It provides explicit type information for TypeScript
- It helps catch errors earlier during development


Without the type specification, TypeScript will infer the type from `data.state.product`, but explicitly declaring it ensures type safety throughout the component.

src/routes/products/[barcode]/edit/+page.svelte
Comment (swastikCommits) [reply src/routes/products/[barcode]/edit/+page.svelte]: Done 👍

VaiTon:
Please rebase on main. I fixed a null issue that prevents the vercel preview from working

swastikCommits:
Done @VaiTon 👍

VaiTon (approved):

PR #293, opened 2025-03-16
added preferences in card

### LINKED ISSUE #282: Add preferences in card of product
The card as of now has name, image, brand and amount, but I was thinking of adding a preference in the card as well like
before

Image
After

Image

This will give user better understanding of the product without actually opening it
And I was also thinking about changing the product card from vertical to horizontal

what are your views on this ? if you are ok with it, then I would like to work on this issue

Bhavishya2601:
### What
Added preference in the cards, so that user can get more info without opening it

there was a small bug in the eco score on the product page, I solved that as well

### Screenshot
image

### Fixes bug(s)
Fixes #282


src/lib/ui/SmallProductCard.svelte
```diff
@@ -1,17 +1,23 @@
...
 	import type { ProductReduced } from '$lib/api';
 	import { navigating } from '$app/state';
+	import { KP_ATTRIBUTE_IMG } from '$lib/const';
 
 	interface Props {
 		product: ProductReduced;
 	}
 
 	let { product }: Props = $props();
+	console.log(product)
```
VaiTon, line 11: Remove this debug statement

src/lib/ui/SmallProductCard.svelte
```diff
@@ -1,17 +1,23 @@
...
+	let nutriscore_src = KP_ATTRIBUTE_IMG('nutriscore-' + (product.nutriscore_grade ?? 'unknown') +'-new-en.svg')
+	let nova_src = KP_ATTRIBUTE_IMG('nova-group-' + (product.nova_group ?? 'unknown') + '.svg')
+	let ecoscore_src = KP_ATTRIBUTE_IMG('ecoscore-' + (product.ecoscore_grade ?? 'unknown') + '.svg')
```
VaiTon, line 14: You should make these three props reactive by using `$derived`. Also, I don't like the `??` operator here, please create another prop. Last thing, you should use the `camelCase` for var names.

src/lib/ui/SmallProductCard.svelte
```diff
@@ -1,17 +1,23 @@
...
+
 </script>
 
 <a
```
VaiTon, line 18: Can we make the the ecoscore / nutriscore stay only under the text? I'd like the image to take the entire height.

VaiTon (changes requested):

VaiTon:
Also, please run `pnpm format` always before committing :)

src/lib/api/product.ts
```diff
@@ -283,7 +283,10 @@ const REDUCED_FIELDS = [
 	'code',
 	'product_name',
 	'brands',
-	'quantity'
+	'quantity',
+	'nutriscore_grade',
+	'ecoscore_grade',
+	'nova_group'
```
teolemon, line 289: This should be conditional on product/project_type

src/lib/ui/SmallProductCard.svelte
```diff
@@ -1,17 +1,30 @@
...
+		KP_ATTRIBUTE_IMG('nutriscore-' + product.nutriscore_grade + '-new-en.svg')
+	);
+
+	let novaSrc = $derived(
+		product.nova_group
+			? KP_ATTRIBUTE_IMG('nova-group-' + product.nova_group + '.svg')
+			: KP_ATTRIBUTE_IMG('nova-group-unknown.svg')
+	);
+
+	let ecoscoreSrc = $derived(KP_ATTRIBUTE_IMG('ecoscore-' + product.ecoscore_grade + '.svg'));
```
teolemon, line 22: This should be conditional on product/project_type

src/lib/ui/SmallProductCard.svelte
```diff
@@ -33,6 +46,11 @@
 			<p class="mt-2 text-sm font-light">
 				{product.brands} - {product.quantity}
 			</p>
+			<div class="mt-2 flex flex-row items-center justify-between gap-2">
+				<div><img src={nutriscoreSrc} alt="nutriscore" class="h-8" /></div>
+				<div><img src={novaSrc} alt="nova" class="h-8" /></div>
+				<div><img src={ecoscoreSrc} alt="nova" class="h-8" /></div>
+			</div>
```
teolemon, line 53: This should be conditional on product/project_type

src/lib/api/product.ts
```diff
@@ -85,16 +85,32 @@ export class ProductsApi {
 		return res.status === 200;
 	}
 
-	async getProductReducedForCard(barcode: string): Promise<ProductState<ProductReduced>> {
+	async getProductReducedForCard(barcode: string, product_type: string): Promise<ProductState<ProductReduced & { product_type: string }>> {
```
VaiTon, line 88: That's not a type we can return. It's too messy. Either we move the responsibility of we clarify better what does this method do.

Too many constants everywhere. We can't keep using `"food"`. It's too error prone and not typed.

VaiTon (changes requested):

src/lib/api/product.ts
Bhavishya2601: I made the changes according to your previous review, but later teolemon asked that the preferences should be displayed conditionally based on product_type, so that they appear only if the product is food otherwise not.
After reviewing the data, I couldn't find any field about the product_type, so I had to create one myself despite knowing it's not the best practice.
 
 But I think having them conditional is of no use as if you go to OFF site, and search for some product other than food like shampoo, it will show results with the preferences
 
image

If you are ok with this, then I can remove the last commit or if you have any suggestions, you can let me know

VaiTon: I think we need to create a better type system, in which we can discriminate with product types from different databases. Maybe this requires one or two another PRs, but it's worth the work.

VaiTon:
Let's keep this WIP and work on the type system

Bhavishya2601:
I have done it for OFF, as for OBF, OPF and rest of them, we can do it after issue #113 is completed

VaiTon:
Please run `pnpm format`

Bhavishya2601:
sry, will remember to run it from next time

VaiTon (approved):

Issue #216, opened 2025-03-06
Dropdown menu for selecting country overflows on smaller screens.


Prathamesh0901:
### What

When we go to mobile view the country options dropdown overflows.

![Image](https://github.com/user-attachments/assets/09779867-e984-496e-b036-4f6af622df15)

### Steps to reproduce the behavior:
1. Go to settings on a mobile screen.
2. Click on country.

### Expected behavior

1. The country dropdown should not overflow.

Prathamesh0901:
i am open to work on this issue.

VaiTon:
Please, please, please, only open an issue on a SINGLE problem.

> When we go to mobile view the country options dropdown overflows.

Sure, that's an issue. You can work on that if you'd like.

> Also when the query and barcode input box is replaced with search button on smaller screens the option to scan the barcode disappears.

And that's another issue, and another issue needs another GitHub issue. Please open another one.

Also, please use a descriptive title. "UI not responsive" is not one.

PR #296, opened 2025-03-17
fix: country dropdown menu overflowing horizontally

swastikCommits:
### What
Fixed the drop down menu for Country options overflowing horizontally in Settings page.

### Screenshot
Screenshot 2025-03-17 162921

### Fixes bug(s)
#216 

### Part of 
None


VaiTon:
I mean.... I really liked the other part... Can't we display it with an ellipsis or something like that?

swastikCommits:
> I mean.... I really liked the other part... Can't we display it with an ellipsis or something like that?

Yes it was tho, but we must also keep our website responsive to all screens too :)

VaiTon (approved):
Argh... I will find a way to do this xD

Issue #181, opened 2025-03-01
Bring features from the Power User Script


VaiTon:

swastikCommits:
reference to #305 @VaiTon 

https://github.com/user-attachments/assets/25799043-7708-4c44-80e2-5f7e2e0bdb84

VaiTon:
Closing in favour of #657

PR #298, opened 2025-03-17
feat: implementation of rotation of images for logged in users

swastikCommits:
### What
In both components:
- Added rotation state and 90-degree rotate function.
- Restricted to logged-in users via preferences store.
- Applied smooth CSS rotation animation.

In ImageButton.svelte:
- Small, semi-transparent circular rotate button in bottom-right, solid on hover, for logged-in users only.

In ImageModal.svelte:
- Larger, semi-transparent circular rotate button in bottom-right, solid on hover, for logged-in users only.
- Resets rotation on modal open/close.

### Screen Recording

https://github.com/user-attachments/assets/b933aabf-14c1-4b69-a638-34d7f3a9c99a

### Part of 
#181

### DIFF
--- a/src/lib/ui/ImageButton.svelte
+++ b/src/lib/ui/ImageButton.svelte
@@ -10,10 +10,9 @@
 	let onclick = $derived(src != null ? () => modal.displayImage(src, alt) : undefined);
 
 	let rotation = $state(0);
-	let isLoggedIn = $derived(get(preferences).username != null && get(preferences).password != null);
 
 	function rotateImage() {
-		rotation = (rotation + 90) % 360;
+		rotation = rotation + 90;
 	}
 </script>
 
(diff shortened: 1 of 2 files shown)


VaiTon:
Why should we restrict this feature to logged in users?

VaiTon:
Okay, so that was for rotating them "permanently" (via API call), and that needs an account. 
What you're doing here is just doing it in the frontend, which is fine by me, but we should not restrict it to logged in users.

src/lib/ui/ImageButton.svelte
```diff
@@ -1,15 +1,44 @@
...
 	let { src, alt }: Props = $props();
 
 	let modal: ImageModal;
 	let onclick = $derived(src != null ? () => modal.displayImage(src, alt) : undefined);
+
+	let rotation = $state(0);
+	let isLoggedIn = $derived(get(preferences).username != null && get(preferences).password != null);
+
+	function rotateImage() {
+		rotation = (rotation + 90) % 360;
```
VaiTon, line 16: Don't circle back on 360, otherwise the animation is not good. Just keep increasing it.

swastikCommits:
> Okay, so that was for rotating them "permanently" (via API call), and that needs an account. What you're doing here is just doing it in the frontend, which is fine by me, but we should not restrict it to logged in users.

So, shall I make changes such that anyone can rotate the images as I didn't use API call to rotate it?

swastikCommits:
@VaiTon I have removed and re-implemented the logic to rotate images to all users

VaiTon (approved):

Issue #297, opened 2025-03-17
chore: lint errors in `pnpm lint` while all the files are already in prettier code style


swastikCommits:
### What
35 errors in `pnpm lint`

```
ASUS@DESKTOP-R7CTGHT MINGW64 ~/Desktop/Coding/OSS/openfoodfacts-explorer (main)
$ pnpm lint

> openfoodfacts-sveltekit@0.0.1 lint C:\Users\ASUS\Desktop\Coding\OSS\openfoodfacts-explorer
> prettier --check . && eslint .

Checking formatting...
All matched files use Prettier code style!

C:\Users\ASUS\Desktop\Coding\OSS\openfoodfacts-explorer\src\lib\knowledgepanels\Element.svelte
  50:7  error  Each block should have a key  svelte/require-each-key
  56:6  error  Each block should have a key  svelte/require-each-key
  59:8  error  Each block should have a key  svelte/require-each-key

C:\Users\ASUS\Desktop\Coding\OSS\openfoodfacts-explorer\src\lib\knowledgepanels\Panel.svelte
  23:2  error  Each block should have a key  svelte/require-each-key

C:\Users\ASUS\Desktop\Coding\OSS\openfoodfacts-explorer\src\lib\knowledgepanels\PanelGroup.svelte
  20:3  error  Each block should have a key  svelte/require-each-key

C:\Users\ASUS\Desktop\Coding\OSS\openfoodfacts-explorer\src\lib\knowledgepanels\Panels.svelte
  28:3  error  Each block should have a key  svelte/require-each-key
  40:1  error  Each block should have a key  svelte/require-each-key

C:\Users\ASUS\Desktop\Coding\OSS\openfoodfacts-explorer\src\lib\ui\ImageModal.svelte
  36:20  error  svelte-ignore comment is used, but not warned  svelte/no-unused-svelte-ignore
  37:20  error  svelte-ignore comment is used, but not warned  svelte/no-unused-svelte-ignore

C:\Users\ASUS\Desktop\Coding\OSS\openfoodfacts-explorer\src\routes\+error.svelte
  33:6  error  Each block should have a key  svelte/require-each-key

C:\Users\ASUS\Desktop\Coding\OSS\openfoodfacts-explorer\src\routes\folksonomy\+page.svelte
  14:2  error  Each block should have a key  svelte/require-each-key

[...]

PR #299, opened 2025-03-17
chore: fixes some pnpm lint errors

swastikCommits:
### What
Fixed all the 35 errors in `pnpm lint`

### Screenshot
Screenshot 2025-03-17 213242

### Fixes bug(s)
#297 

### Part of 
None


swastikCommits:
@teolemon #277  added these error(shown below) while fixed the previous 35 errors.
Can you review this PR? 
imo this error can be fixed from `labeler.yml`, if you want then I can push a commit fixing this error too.

```
Error:  .github/labeler.yml: SyntaxError: All collection items must start at the same column (30:5)
Error:    28 |       - any-glob-to-any-file: 'src/routes/products/[barcode]/edit/TagsString.svelte'
Error:    29 | 💸 Prices:
Error:  > 30 |   - changed-files:
Error:       |     ^^^^^^^^^^^^^^
Error:  > 31 |       - any-glob-to-any-file: 'src/routes/products/[barcode]/Prices.svelte'
Error:       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error:  > 32 |       src/lib/api/prices.ts
Error:       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error:  > 33 | 🖼️ Images:
Error:       | ^
Error:    34 |   - changed-files:
Error:    35 |       - any-glob-to-any-file: 'src/lib/ui/ImageModal.svelte'
Error:    36 |       - any-glob-to-any-file: 'src/lib/ui/ImageButton.svelte'
All matched files use Prettier code style!
 ELIFECYCLE  Command failed with exit code 2.
Error: Process completed with exit code 2.
```

src/routes/+error.svelte
```diff
@@ -30,7 +30,7 @@
 				</p>
 
 				<ul class="list-disc">
-					{#each errors as error}
+					{#each errors as error (error.message)}
```
VaiTon, line 33: Please use the error index as key here

src/routes/products/[barcode]/+page.svelte
```diff
@@ -70,8 +70,8 @@
 				{#await data.taxo.brands}
 					Loading...
 				{:then brands}
-					{#each product.brands_tags as tag, i}
```
Comment (VaiTon) [inline src/routes/products/[barcode]/+page.svelte:73]: What was wrong with this `if i > 0`?

src/routes/products/[barcode]/+page.svelte
```diff
@@ -98,8 +98,8 @@
 				{#await data.taxo.stores}
 					Loading...
 				{:then stores}
-					{#each product.stores_tags as tag, i}
-						{#if i > 0},
```
Comment (VaiTon) [inline src/routes/products/[barcode]/+page.svelte:102]: Same here

src/routes/products/[barcode]/+page.svelte
```diff
@@ -111,8 +111,8 @@
 				{#await data.taxo.labels}
 					Loading...
 				{:then labels}
-					{#each product.labels_tags as tag, i}
-						{#if i > 0},
```
Comment (VaiTon) [inline src/routes/products/[barcode]/+page.svelte:115]: Same here

src/routes/products/[barcode]/+page.svelte
```diff
@@ -141,8 +141,8 @@
 				{#await data.taxo.origins}
 					Loading...
 				{:then origins}
-					{#each product.origins_tags as tag, i}
-						{#if i > 0},
```
Comment (VaiTon) [inline src/routes/products/[barcode]/+page.svelte:145]: Same here

src/routes/products/[barcode]/Prices.svelte
```diff
@@ -27,70 +27,95 @@
 		nearStores = await getNearStores();
 	});
 
+	// Common currencies that are definitely supported by the API
+	type Currency = 'EUR' | 'USD' | 'GBP' | 'JPY' | 'CNY' | 'INR' | 'AUD' | 'CAD' | 'CHF' | 'HKD';
```
Comment (VaiTon) [inline src/routes/products/[barcode]/Prices.svelte:31]: This should not be defined in a .svelte

src/routes/products/[barcode]/Prices.svelte
```diff
@@ -27,70 +27,95 @@
...
+
 	let newPrice = $state({
```
Comment (VaiTon) [inline src/routes/products/[barcode]/Prices.svelte:33]: As we're defining a type for each property, it would be better to define it for the entire var

src/routes/products/[barcode]/Prices.svelte
```diff
@@ -27,70 +27,95 @@
...
 
 	async function login() {
 		console.debug('Logging in...');
-		const res = await pricesApi.login({
-			username: loginFields.email,
-			password: loginFields.password
-		});
-		if (res.error != null) {
-			console.error('Error while logging in', res.error);
+		try {
```
Comment (VaiTon) [inline src/routes/products/[barcode]/Prices.svelte:46]: Try to keep `try` block the smallest possible. Here, you don't need to try the if block.

src/routes/products/[barcode]/Prices.svelte
```diff
@@ -27,70 +27,95 @@
...
-			location_osm_type: type as 'NODE' | 'WAY' | 'RELATION'
-		});
-
-		if (res.error != null) {
-			console.error('Error while submitting price', res.error);
-		} else {
-			console.debug('Submitted price', res.data);
-			prices.items.push(res.data);
-			invalidateAll();
+		try {
```
Comment (VaiTon) [inline src/routes/products/[barcode]/Prices.svelte:83]: Same here for the try

src/routes/products/[barcode]/Prices.svelte
```diff
@@ -101,25 +126,27 @@
...
 						<td>
-							{#await idToName(fetch, price.location_osm_id)}
+							{#await idToName(fetch, price.location_osm_id ?? 0)}
 								Loading...
 							{:then storeName}
 								<a
-									href={`https://www.openstreetmap.org/${price.location_osm_type.toLowerCase()}/${
-										price.location_osm_id
+									href={`https://www.openstreetmap.org/${(price.location_osm_type ?? 'node').toLowerCase()}/${
+										price.location_osm_id ?? 0
```
Comment (VaiTon) [inline src/routes/products/[barcode]/Prices.svelte:138]: It would not make sense to show a link to a /0. We should not display the map if that happens

src/routes/products/[barcode]/Prices.svelte
```diff
@@ -101,25 +126,27 @@
...
 									}`}
 								>
 									{storeName}
 								</a>
 							{:catch error}
-								<span class="text-red-500">Error: {error.message}</span>
+								<span class="text-red-500"
+									>Error: {error instanceof Error ? error.message : String(error)}</span
```
Comment (VaiTon) [inline src/routes/products/[barcode]/Prices.svelte:145]:
```suggestion
									>Error: {error instanceof Error ? error.message : error}</span
```

src/routes/products/[barcode]/Prices.svelte
```diff
@@ -101,25 +126,27 @@
...
+								>
 							{/await}
 						</td>
-						<td>{new Date(price.date).toLocaleDateString()}</td>
+						<td>{new Date(price.date ?? '').toLocaleDateString()}</td>
```
Comment (VaiTon) [inline src/routes/products/[barcode]/Prices.svelte:149]: If this happens we should display a "no date" label, not format a date

VaiTon (changes requested):
Thank you for the PR! 

This is the first round of comments. As this PR spannes multiple files, expect another after the changes.

src/routes/products/[barcode]/Prices.svelte
Comment (swastikCommits) [reply src/routes/products/[barcode]/Prices.svelte:33]: 1. Created a `NewPrice` interface in `currency.ts` that properly defines the shape of the newPrice state variable
2. Added a `OsmType` type to better type the `location_osm_type` field
3. Updated the Svelte component to import `NewPrice` and explicitly type the `newPrice` state variable with `NewPrice`

src/routes/products/[barcode]/Prices.svelte
Comment (swastikCommits) [reply src/routes/products/[barcode]/Prices.svelte:138]: 1. Added a conditional check `{#if price.location_osm_id && price.location_osm_id !== 0}` to verify both that the ID exists and is not 0
2. Removed the fallback `?? 0` in the URL since we're already checking that the ID exists
3. Added an` {:else}` block to display just the store name without a link when the ID is missing or 0

src/routes/products/[barcode]/+page.svelte
Comment (VaiTon) [reply src/routes/products/[barcode]/+page.svelte:73]: Why is this code problematic? I understand the change you made, but I don't understand why the old check was not good enough

Comment (swastikCommits) [reply src/routes/products/[barcode]/+page.svelte:73]: Omitting a key in Svelte list rendering causes issues when the list changes, as Svelte can't track items properly, leading to inefficient updates or lost state. 
Adding a key (like index `i`) improves DOM efficiency, preserves state, and aids animations. 
Index as a key is okay for static lists but less ideal for reorderable one, so unique IDs are best when available.

Comment (VaiTon) [reply src/routes/products/[barcode]/+page.svelte:73]: That's not what I asked... I asked why did you change the `i > 0` condition...

VaiTon (changes requested):

VaiTon:
Can we split this into multiple PRs, so that we can merge some of them? Otherwise I have to block the entire PR because of some change on a single file that I don't like

src/routes/products/[barcode]/+page.svelte
Comment (swastikCommits) [reply src/routes/products/[barcode]/+page.svelte:73]: Sir actually both of them does the same thing. Earlier for `i>0` the logic was it looks for comma before all the items, while `i < product.brands_tags.length - 1` checks for comma after the items.
Looking for commas after for example Brand1, Brand2 would be more conventional imo.

But @VaiTon if you want I can revert back to use the same logic as used earlier.

swastikCommits:
> Can we split this into multiple PRs, so that we can merge some of them? Otherwise I have to block the entire PR because of some change on a single file that I don't like

If so, @VaiTon can you point out what all files shall I separate using a different PR? or else I'm here to resolve all the conversations as you suggest to push all the files at a go.

src/routes/products/[barcode]/+page.svelte
Comment (swastikCommits) [reply src/routes/products/[barcode]/+page.svelte:102]: Sir actually both of them does the same thing. Earlier for i>0 the logic was it looks for comma before all the items, while i < product.brands_tags.length - 1 checks for comma after the items.
Looking for commas after for example Brand1, Brand2 would be more conventional imo.

src/routes/products/[barcode]/+page.svelte
Comment (swastikCommits) [reply src/routes/products/[barcode]/+page.svelte:115]: Sir actually both of them does the same thing. Earlier for i>0 the logic was it looks for comma before all the items, while i < product.brands_tags.length - 1 checks for comma after the items.
Looking for commas after for example Brand1, Brand2 would be more conventional imo.

src/routes/products/[barcode]/+page.svelte
Comment (VaiTon) [reply src/routes/products/[barcode]/+page.svelte:73]: Please revert and then I'll merge

Comment (swastikCommits) [reply src/routes/products/[barcode]/+page.svelte:73]: Ok @VaiTon undoing the changes

swastikCommits:
After new merges, we are still getting 17 lint errors under `pnpm lint`
I would probably try to raise another PR fixing those too.

VaiTon:
Yes, I reverted some files to be able to merge the PR. Feel free to open a new PR!

PR #305, opened 2025-03-18
feat: implemented real-time Energy Calculation in edit page

swastikCommits:
### What
Created an `Energy.svelte` component that:

- Accepts nutriments data as input
- Provides input fields for protein, carbs, and fat
- Calculates energy in real-time (4 kcal/g protein & carbs, 9 kcal/g fat)
- Displays energy in kcal & kJ (1 kcal = 4.184 kJ)
- Auto-updates nutriments data on change
- Added to the product edit page between product info and ingredients sections

### Screen Recordings

https://github.com/user-attachments/assets/6336800c-5bd3-427d-98e3-6f85ba0af16c

### Part of 
#181


src/routes/products/[barcode]/edit/Energy.svelte
```diff
@@ -0,0 +1,72 @@
...
+	let proteins = $state(nutriments.proteins || 0);
+	let carbohydrates = $state(nutriments.carbohydrates || 0);
+	let fat = $state(nutriments.fat || 0);
+
+	let calculatedEnergy = $derived(proteins * 4 + carbohydrates * 4 + fat * 9);
+
+	// Calculate energy in kJ (1 kcal = 4.184 kJ)
+	let calculatedEnergyKj = $derived(calculatedEnergy * 4.184);
+
+	$effect(() => {
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/Energy.svelte:20]: I don't understand what you're trying to do here, but this is 99% not the way. If you're trying to make the child modify some parent values (two way binding) you should use the `bind` construct:

https://svelte.dev/docs/svelte/bind

VaiTon (changes requested):

src/routes/products/[barcode]/edit/Energy.svelte
Comment (swastikCommits) [reply src/routes/products/[barcode]/edit/Energy.svelte:20]: The code is doing data derivation and synchronization in a Svelte 5 way using the new `$state`, `$derived`, and` $effect` syntax.


This is actually a valid approach, let me explain what's happening:
1. The component receives a `nutriments` object as a prop
2. It extracts values into local reactive state with` $state`
3. It calculates derived values with `$derived`
4. It uses `$effect` to synchronize changes back to the parent object

We can use two-way binding, but this will look different. The current approach separates the 
Local state management, Derived calculations and Syncing back to the parent. But then you'd need to handle the derived calculations differently since they depend on these values changing.

Comment (VaiTon) [reply src/routes/products/[barcode]/edit/Energy.svelte:20]: Then this process should not be handled in a sub-component. I don't really see (the screen recording is not working) how this works. The issue is only about creating an alert if values do not match (I guess, @teolemon created it so only he can be sure).

Comment (swastikCommits) [reply src/routes/products/[barcode]/edit/Energy.svelte:20]: @VaiTon I'm uploading another instance of this screen recording in the corresponding issue page too. Have a look if it works there

Comment (swastikCommits) [reply src/routes/products/[barcode]/edit/Energy.svelte:20]: @VaiTon if you were able to see the screen recording I tagged, then can you explain me what exactly changes do we want here?
In default it shows the Total Energy(kcal and kJ) of that products as per their macros, further we can also make changes to those macros(protein, carbs, fat) to render the new energy calculated.

swastikCommits:
@VaiTon any changes required here? 
I have pinned the screen recording here in #181

VaiTon:
@teolemon is this what you meant with " Edit mode: compute and display energy in real-time" in #181 ?

swastikCommits:
@teolemon got it will try to replicate like the one from the main website

github-throwaway:
Are you opening a numeric keyboard on mobile for these fields? Design looks good but I think @teolemon might be more opinionated than me 😄

swastikCommits:
Yeah, wby @teolemon and @VaiTon ?

VaiTon:
@swastikCommits good looking for me, now I need the code

swastikCommits:
Thank you for the update. I will push the code as soon as possible.

VaiTon:
Stale PR. Closing. Feel free to reopen!

Issue #306, opened 2025-03-18
chore(readme): Add `.env` Setup Instructions to README for Proper development Configuration


swastikCommits:
### What
```
SvelteKitError: Not found: /keys
    at resolve (C:\Users\ASUS\Desktop\Coding\openfoodfacts-explorer-2\openfoodfacts-explorer\node_modules\.pnpm\@sveltejs+kit@2.19.0_@svelt_11e4dce83cfa5e48bd9c59c6830f6df7\node_modules\@sveltejs\kit\src\runtime\server\respond.js:531:13)
    at resolve (C:\Users\ASUS\Desktop\Coding\openfoodfacts-explorer-2\openfoodfacts-explorer\node_modules\.pnpm\@sveltejs+kit@2.19.0_@svelt_11e4dce83cfa5e48bd9c59c6830f6df7\node_modules\@sveltejs\kit\src\runtime\server\respond.js:321:5)
    at init_promise.#options.hooks.handle (C:\Users\ASUS\Desktop\Coding\openfoodfacts-explorer-2\openfoodfacts-explorer\node_modules\.pnpm\@sveltejs+kit@2.19.0_@svelt_11e4dce83cfa5e48bd9c59c6830f6df7\node_modules\@sveltejs\kit\src\runtime\server\index.js:76:56)
    at respond (C:\Users\ASUS\Desktop\Coding\openfoodfacts-explorer-2\openfoodfacts-explorer\node_modules\.pnpm\@sveltejs+kit@2.19.0_@svelt_11e4dce83cfa5e48bd9c59c6830f6df7\node_modules\@sveltejs\kit\src\runtime\server\respond.js:318:40)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async file:///C:/Users/ASUS/Desktop/Coding/openfoodfacts-explorer-2/openfoodfacts-explorer/node_modules/.pnpm/@sveltejs+kit@2.19.0_@svelt_11e4dce83cfa5e48bd9c59c6830f6df7/node_modules/@sveltejs/kit/src/exports/vite/dev/index.js:547:22 {
  status: 404,
  text: 'Not Found'
```
This error while developing may arise to developers in the terminal window.
It won't enable the developers to access /folksonomy page while in development mode.

### Steps to reproduce the behavior:
1. Clone and install the dependencies
2. Run `pnpm run dev` and go to /folksonomy route
3. This will most probably show an error in the terminal: `Not found /keys`

[...]

PR #307, opened 2025-03-18
chore: created `.env.example` for environment variables

swastikCommits:
### What
Updates the `README.md` to include instructions for setting up the `.env` file. Developers can now easily configure their environment by copying `.env.example` and renaming it to ` .env`

1. Added a "Environment Setup" section to README.md
2. Provided a command for copying `.env.example` to `.env
`

### Fixes bug(s)
#306

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+VITE_FOLKSONOMY_API_URL=https://api.folksonomy.openfoodfacts.org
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -0,0 +1 @@
+VITE_FOLKSONOMY_API_URL=https://api.folksonomy.openfoodfacts.org
```
VaiTon, line 1: Please point to the staging env, as we don't want devs making changes to the actual folksonomy db while working

VaiTon (changes requested):

.env.example
swastikCommits: Done 👍

PR #310, opened 2025-03-18
delete language in product edit page and show the languages in add language section

### LINKED ISSUE #309: Delete language in product edit page
### Problem
Add a feature to delete a language in products edit page
The reason behind is I was exploring the add the language section and clicked randomly on some languages and found that there's no way to actually delete them

### Proposed solution
Here we can first show all the added languages in the field and they can have a cross sign next to it, clicking it would delete that specific language
Image

Bhavishya2601:
### What
Add the delete language feature where users can delete a language and I have implemented to show all the languages in the add language section and if while deleting if the languages count is equal to 1, then it won't delete the language (will only delete if language count is more than 1)

### Screenshot
image

### Fixes bug(s)
Fixes #309

### DIFF
--- a/src/routes/products/[barcode]/edit/+page.svelte
+++ b/src/routes/products/[barcode]/edit/+page.svelte
@@ -98,10 +98,38 @@
 	<input type="checkbox" />
 	<div class="collapse-title font-semibold">Add a language</div>
 	<div class="collapse-content text-sm">
-		<label class="input w-full">
+		<div class="mb-4 flex flex-wrap gap-2">
+			{#each Object.keys($productStore.languages_codes) as code}
+				<div class="bg-base-100 flex items-center gap-3 rounded-xl p-2">
+					<span>{getLanguage(code)}</span>
+					<button
+						class="flex items-center justify-center"
+						aria-label="Remove language"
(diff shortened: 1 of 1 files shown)


Bhavishya2601:
@Abbas-Askari  When I was exploring the add a language section, I accidentally clicked randomly and found that there's no way to delete them, so that's why I proposed this idea.

So according to what you said, even adding a new language won't add it ?

Abbas-Askari:
> So according to what you said, even adding a new language won't add it ?

Adding a language to `languages_codes` won't add it. You need to add a field like `ingredients_test_es` to add Spanish for example.

Bhavishya2601:
>Adding a language to languages_codes won't add it. You need to add a field like ingredients_test_es to add Spanish for example.

So right now 'add a language' section is not working, it is kind off for show only

VaiTon:
> it is kind off for show only

Yeah, that's exactly the point. Adding the language via that UI makes it possible to add a field in the specified language. Submitting a product after having created a language but before adding a new field in that language does NOT add the new language to the product. 

I don't really know if we can delete a language, so for now let's close this one. Sorry!

Issue #239, opened 2025-03-09
Location marker of product processing place on map not working on product details page


Prathamesh0901:
### What
When we go to product details page by clicking on the products showcased on homepage of website, there is a map down the details which shows where it is manufactured on a world map.
The marker for that feature is not rendering for any of the products.

### Steps to reproduce the behavior:
1. Click on any product on home page.
2. Scroll down to the map.

### Expected behavior
The marker should be visible making it easier for the user to identify it's origin.

### Screenshots

![Image](https://github.com/user-attachments/assets/4bca82d2-4ddf-4273-96de-0dba4aa33ea7)

Prathamesh0901:
i am open to work on this issue.

Prathamesh0901:
hello @VaiTon, I checked the same on my local environment but it is actually working find
![Image](https://github.com/user-attachments/assets/514e3548-4e2f-4c3a-96fa-5750efca5bf8)
honestly, I don't know what might be the issue, if it is working on local but not on production.
what should I do?

swastikCommits:
@Prathamesh0901 its working fine for me too, it would be better to close this issue(if this issue doesn't persist anymore)

swastikCommits:
@Prathamesh0901 please close this issue as completed

PR #313, opened 2025-03-19
fix: location marker gets broken on map

swastikCommits:
### What

- Validated map pointers for correct latitude/longitude.
- Added error handling for marker creation and map setup.
- Ensured cleanup on unmount with proper marker tracking.
- Used `bind:this` for direct map container reference.
- Fixed async code in `onMount`.

These changes prevent marker issues by validating data, handling errors, and avoiding leaks.

### Screenshot
Screenshot 2025-03-19 064135

### Fixes bug(s)
#239

### DIFF
--- a/src/lib/knowledgepanels/Map.svelte
+++ b/src/lib/knowledgepanels/Map.svelte
@@ -9,22 +9,71 @@
 	const TILES_BASE_URL = 'https://tile.openstreetmap.org/{z}/{x}/{y}.png';
 	const ATTRIBUTION = '&copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a>';
 
-	onMount(async () => {
-		const { default: L } = await import('leaflet');
+	let mapContainer: HTMLElement;
+	let mapInstance: any;
+	let markers: any[] = [];
 
-		const map = L.map('map', {});
-		map.fitBounds(
-			element.map_element.pointers.map((pointer) => [pointer.geo.lat, pointer.geo.lng])
-		);
-
-		map.setZoom(Math.min(map.getZoom(), MAX_INITIAL_ZOOM));
-		L.tileLayer(TILES_BASE_URL, { maxZoom: MAX_ZOOM, attribution: ATTRIBUTION }).addTo(map);
-
-		for (const pointer of element.map_element.pointers) {
-			const marker = L.marker([pointer.geo.lat, pointer.geo.lng]);
-			marker.addTo(map);
+	onMount(() => {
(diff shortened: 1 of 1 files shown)


src/lib/knowledgepanels/Map.svelte
```diff
@@ -9,22 +9,71 @@
...
+	let mapContainer: HTMLElement;
+	let mapInstance: any;
+	let markers: any[] = [];
 
-		const map = L.map('map', {});
-		map.fitBounds(
-			element.map_element.pointers.map((pointer) => [pointer.geo.lat, pointer.geo.lng])
-		);
+	onMount(() => {
+		if (!element?.map_element?.pointers?.length) {
```
VaiTon, line 17: Using `element`, which is a prop, inside the onMount is a problem. If the prop changes the onMount is not called again. Every kind of logic referencing props should be done inside reactive runes, like $effect and $derived

VaiTon (changes requested):

swastikCommits:
@VaiTon have a look I refactored and the changes are the following: 
- Using `$effect` to update the map when element changes.
- Splitting map setup (`onMount`) from data updates (`updateMap`).
- Adding cleanup for markers and map.
- Fixing async handling in `onMount`.
This ensures the map updates dynamically with new prop data, not just on mount.

src/lib/knowledgepanels/Map.svelte
```diff
@@ -9,22 +9,100 @@
...
 
-		const map = L.map('map', {});
-		map.fitBounds(
-			element.map_element.pointers.map((pointer) => [pointer.geo.lat, pointer.geo.lng])
-		);
+	// Initialize map on mount
+	onMount(() => {
+		let mounted = true;
+
+		(async () => {
```
VaiTon, line 21: Why this `async`?

swastikCommits: 1. The primary reason is to allow for using `await` with the dynamic import of the Leaflet library, this loads the Leaflet library only when needed (code splitting) rather than bundling it with the initial page load.
2. The `mounted` flag helps prevent attempting to update the component after it's been unmounted, which could happen if the asynchronous import completes after navigation away from the component.
3. It keeps the asynchronous code contained while still allowing the `onMount` function to return its cleanup function properly.

VaiTon: Seems reasonable. Please add it as a comment for the future me xD

VaiTon (approved):

PR #321, opened 2025-03-19
Improved search results with full product details

sohan-gupthak:
## Current Behavior
The search functionality currently loads minimal product data from the search API, which can lead to incomplete information when navigating to product details pages. Additionally, there were type safety issues in handling product states.

## Proposed Changes
implemented the following improvements:

### Data Loading
- Modified the search loader to fetch full product details for each search result
- Added proper error handling for failed product detail requests
- Maintains pagination while loading complete product information

### Type Safety Improvements
- Added proper TypeScript types for search results and product states
- Implemented type guards for safe product state handling
- Created a `ProductSearchResult` type to properly type the search response

### Code Structure
```typescript
type ProductSearchResult = {
    count: number;
    page: number;
    page_size: number;
    page_count: number;
    products: ProductStateFound<Product>[];
    total_pages: number;
};

### DIFF
--- a/src/routes/products/search/+page.svelte
+++ b/src/routes/products/search/+page.svelte
@@ -2,9 +2,25 @@
 	import { page } from '$app/state';
 	import SmallProductCard from '$lib/ui/SmallProductCard.svelte';
 	import type { PageData } from './$types';
+	import type { Product, ProductState, ProductStateFound } from '$lib/api/product';
+
+	type ProductSearchResult = {
+		count: number;
+		page: number;
+		page_size: number;
+		page_count: number;
+		products: ProductStateFound<Product>[];
+		total_pages: number;
+	};
+
+	function isSuccessState(state: ProductStateFound<Product>): state is ProductStateFound<Product> {
+		return state.status === 'success' || state.status === 'success_with_warnings';
+	}
 
 	interface Props {
-		data: PageData;
+		data: PageData & {
+			result: Promise<ProductSearchResult>;
+		};
 	}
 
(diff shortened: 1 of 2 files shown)


sohan-gupthak:
@VaiTon , @teolemon  waiting for review.

VaiTon:
> The search functionality currently loads minimal product data from the search API, 

This is the intended behavior, as we don't need a lot of info there.

> which can lead to incomplete information when navigating to product details pages.

If that's the case, this is the issue. The main product page should load the full product (and I think it already does?)

sohan-gupthak:
> If that's the case, this is the issue. The main product page should load the full product (and I think it already does?)

@VaiTon, Yes it does exits but due to many request at a time I guess I have frequently encountering the below issue.
 
Screenshot 2025-03-20 at 10 49 10 AM

If this is an issue can I work on API's optimization and also when we hover on any product card there are multiple backend requests happening rather we can make backend requests only when we click on a particular product.

VaiTon:
> > If that's the case, this is the issue. The main product page should load the full product (and I think it already does?)
> 
> @VaiTon, Yes it does exits but due to many request at a time I guess I have frequently encountering the below issue.
> 
> Screenshot 2025-03-20 at 10 49 10 AM
> 
> If this is an issue can I work on API's optimization and also when we hover on any product card there are multiple backend requests happening rather we can make backend requests only when we click on a particular product.

Maybe that's because of sveltekit preloading everything. Can you please open an issue about it?

About this PR, I think we should close it for now.

PR #322, opened 2025-03-19
OBF Integration Into Search & Product Page

KarruHarin:
### What
OBF Integration Into Search & Product Page

### Screenshot

https://github.com/user-attachments/assets/8230fa07-92d1-43f7-b6e0-fbdfc018e154

### My Approach
As @teolemon mentioned I removed nova,eco scores for beauty products and as @VaiTon said I made and used a typeContent Object to display data of OBF or OFF  instead of {#if} .

For search I just followed the method used in mobiles as shown by @teolemon ,which is a drop down menu to search in respective databases .

### Fixes bug(s)
#113 

### Next Steps 
I am ready to do required changes or modifications 
and will be adding Edit integration too.

If you think this approach is not scalable I am ready to discuss an alternative approach !

Thank you :)

### DIFF
--- a/src/lib/ui/Logo.svelte
+++ b/src/lib/ui/Logo.svelte
@@ -1,4 +1,4 @@
-<picture>
+<picture class="mx-2 block h-10 w-44 sm:h-12 sm:w-68 lg:mx-6">
 	<source
 		srcset="https://static.openfoodfacts.org/images/logos/off-logo-horizontal-dark.svg"
 		media="(prefers-color-scheme: dark)"
@@ -6,5 +6,6 @@
 	<img
 		src="https://static.openfoodfacts.org/images/logos/off-logo-horizontal-light.svg"
 		alt="OpenFoodFacts Explorer"
+		class="h-full w-full"
 	/>
 </picture>
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -47,7 +47,7 @@
 	<meta name="viewport" content="width=device-width, initial-scale=1" />
 </svelte:head>
 
-<div class="bg-base-100 navbar hidden px-10 lg:flex lg:px-40">
+<div class="bg-base-100 navbar hidden px-10 lg:flex lg:justify-center lg:px-30 xl:px-40">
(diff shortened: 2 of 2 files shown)


src/routes/products/[barcode]/+page.ts
```diff
@@ -16,12 +16,21 @@ import { PricesApi, isConfigured as isPricesConfigured } from '$lib/api/prices';
...
+	let state = await productsApi.getProduct(params.barcode);
+	let type: ProductSourceType = 'OFF';
 	if (state.status === 'failure') {
-		error(404, { message: 'Failure to load product', errors: state.errors });
+		type = 'OBF';
+		const beautyApi = new BeautyApi(fetch);
+		state = await beautyApi.getProduct(params.barcode);
+		if (state.status === 'failure') {
+			error(404, { message: 'Failure to load product', errors: state.errors });
+		}
```
Comment (VaiTon) [inline src/routes/products/[barcode]/+page.ts:33]: @openfoodfacts/openfoodfacts-server  we're doing a trial and error here to get the right db. Don't we have any API that can discriminate based on the barcode in which db we should search. I'm not really updated on the entire "db merge" thing.

Comment (KarruHarin) [reply src/routes/products/[barcode]/+page.ts:33]: I will confirm it from slacks dev channel , but as far as I have seen there is no API that can discriminate it !

Comment (KarruHarin) [reply src/routes/products/[barcode]/+page.ts:33]: Currently there is no API that can discriminate based on the barcode , so either we have to go with this method or search parallely in all databases at once !  I think this methods better than searching parallely !

I have an idea ,theres  a small tweek we can do , 99% of the times users come to this page from the search page or home page .As of now all data on homePage is only fetched from OFF , and in search as we are already know in which database we are searching(dropdown menu) , we can send that info to product page and search only in that database.Then this trial and error will be only happen when someone directly enters the url of the product page in the browser.This can be a solution but I  think it adds a bit of complexity to the code .

So how do I proceed ?

KarruHarin:
@VaiTon @teolemon  If you think this approach is okay , I will add other databses (OPF etc..) .

PR #330, opened 2025-03-21
feat: Add new products

### LINKED ISSUE #308: Feature to Add new Products
### Problem
If the product is missing then there is no option to add new product for the user

Image

### Proposed solution
Implement the add product feature

I don't know if that possible at the moment, can you confirm @VaiTon @teolemon

Bhavishya2601:
Implemented Add product feature, so that users can add a product if they don't find it

Fixes #308


VaiTon:
We should think about having as much as possible in common with the edit mode. I think this can be made only using the edit route, but maybe I'm missing something

src/routes/addProduct/+page.svelte
```diff
@@ -0,0 +1,460 @@
...
+
+	async function handleImageUpload(event: Event, type: 'front' | 'ingredients') {
+		const file = (event.target as HTMLInputElement).files?.[0];
+		if (file) {
+			try {
+				const base64Image = await convertToBase64(file);
+
+				newProduct.update((product) => {
+					if (type === 'front') {
+						product.image_front_url = `data:${file.type};base64,${base64Image.split(',')[1]}`;
```
VaiTon, line 220: I don't think that's the right way to upload an image to the API

Bhavishya2601: Yeah, there's a error in image upload function also, I'm currently working on it

Bhavishya2601:
>We should think about having as much as possible in common with the edit mode.

Yeah, I am thinking of making it as close to the edit page in oFF.org

>I think this can be made only using the edit route

Yeah I am using the same route that is being used for edit

src/routes/addProduct/+page.svelte
Bhavishya2601: Yeah, I am working on this

Bhavishya2601:
I have made further changes, and there some error, I'm encountering while uploading image, I'm currently working on it, would really appreciate if you could provide some suggestions on it

src/routes/addProduct/+page.svelte
VaiTon, line 1: I don't like the fact that we need a new page. We should modify the `/edit` route so that if the product does not exist we create it (that's why the API call is `addOrEdit`)

src/routes/addProduct/+page.svelte
```diff
@@ -0,0 +1,557 @@
...
+		if (user_id && password) {
+			currentUser = {
+				isLoggedIn: true,
+				user_id,
+				password
+			};
+		} else {
+			alert('You must be logged in access this page');
+			goto('/settings');
+		}
```
VaiTon, line 33: The `onMount` function is not the right one for this kind of checks. I guess the best place would be the +page.ts

VaiTon (changes requested):

github-throwaway:
@Bhavishya2601 can you put this in draft mode and mark it as ready once it has reached that state? Thanks. 🙏

src/routes/addProduct/+page.svelte
Bhavishya2601: but we can't be showing the edit route URL while the user will be adding the product

src/routes/addProduct/+page.svelte
Bhavishya2601: yeah, I'll add it in +page.ts

VaiTon (changes requested):
1. I don't like the `addProduct` url. It should be `/<barcode>/add`
2. The `newProduct` is really really bad. Can you read that? Because I really can't.
3. You're using old Svelte 4 syntax (which is what LLM generate). Was this made by you?

Bhavishya2601:
>It should be /<barcode>/add

I can change that

>You're using old Svelte 4 syntax (which is what LLM generate). Was this made by you?

As I'm still learning, so I took some external help, But if you are saying each part of it is made by LLM then I'm respectfully against it, 
If you scroll up, you'll find I was stuck with image upload part for like a week or so, So in the end I took some help, and maybe the LLM changed the syntax

Bhavishya2601:
@VaiTon Pls review this PR and let me know if you need any changes

VaiTon:
> We should think about having as much as possible in common with the edit mode. I think this can be made only using the edit route, but maybe I'm missing something

I really can't understand why we can't use the same `edit` route? Is there anything missing? Why do we need to create an entirely new flow?

VaiTon (changes requested):

Bhavishya2601:
@VaiTon Check out now, I have modified the edit route to add product

The add route is still there, If you are ok with the edit route then I'll remove the add route

Bhavishya2601:
@VaiTon pls review this and let me know if you need any changes

src/routes/products/[barcode]/edit/+page.svelte
```diff
@@ -41,7 +206,11 @@
 	let storeNames = $derived(getNames(data.stores));
 	let originNames = $derived(getNames(data.origins));
 	let countriesNames = $derived(getNames(data.countries));
-	let productStore = $derived(writable<Product>(data.state.product));
+	let productStore = $derived(
+		writable<Product>(
+			data.state.status === 'empty' ? emptyProduct : (data.state.product ?? emptyProduct)
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/+page.svelte:211]: Can we make this a `$derived.by` and use some `if-else`s? This is not so readable

src/routes/products/[barcode]/edit/+page.ts
```diff
@@ -15,10 +14,19 @@ export const load = (async ({ fetch, params }) => {
 	]);
 
 	if (product.status === 'failure') {
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/+page.ts:16]: Are we sure that `failure` means always "product not found"?

src/routes/products/search/+page.svelte
```diff
@@ -28,9 +28,13 @@
 				<SmallProductCard {product} />
 			{/each}
 		{:else}
-			<div class="mt-8 text-center opacity-70">
+			<div class="mt-8 text-center">
 				<p>No products found</p>
 				<p>We couldn't find any products matching your search</p>
+				<a
+					class="btn btn-secondary join-item mt-5 px-10"
+					href="/products/{page.url.searchParams.get('q')}/edit">Add new Product</a
```
VaiTon, line 36: Add a space after and before the inner content and re-format

VaiTon (changes requested):

src/routes/products/[barcode]/edit/+page.ts
Comment (Bhavishya2601) [reply src/routes/products/[barcode]/edit/+page.ts:16]: Yeah sry, I didn't consider if the product is of different type, let me make the changes

src/routes/products/search/+page.svelte
swastikCommits, line 1: Can we implement the component to "Add a new product" in the search page itself, whenever a product is not found?

Bhavishya2601: I didn't get you

Did you mean implement the "add product" functionality in search page

swastikCommits: Yup, whenever there is "no product found" displayed on search page, just adding the component for "Add product" here would work.

Bhavishya2601: VaiTon doesn't want a whole new component for adding a product, he wants to use the edit route in order to add the product, which is quite reasonable

swastikCommits: You didn't get it ig.
On the search page, whenever a product is not found, add a button below it saying "Add the product" which will redirect it to the same logic you used.

Bhavishya2601: That's exactly what this PR is doing

https://github.com/user-attachments/assets/e245bac3-38e6-45f7-a4a6-3ea726c43924

VaiTon: A propos, the "Product not found (create a new product)" div should be centered

swastikCommits: Ah got it, I didn't check the code earlier :(
Btw except on the search page from where all can the user access this feature to add the product?

Bhavishya2601: >Btw except on the search page from where all can the user access this feature to add the product?

Currently it's only on search page with the intention that if the user doesn't find the product, he can add it

Bhavishya2601:
@VaiTon check out now

src/routes/products/search/+page.svelte
swastikCommits: How about adding something like "Wanna be a part of OFF family? ->Add a new product". This can be placed on the homepage(or somewhere else). As of now this feature would be only available in the search page whenever some product is not found in the db. 
Adding a new product is a quite useful feature so, something like that on the homepage will keep the users engaged in the platform as well. This needs further discussion as well @VaiTon

Bhavishya2601: It seems a good idea at first but I don't think it's practical

we are checking if the product exists in db before allowing the user to add it, so by placing the option to add product on homepage would want to first check if the product is present or not, which is what search functionality is about

swastikCommits: Yeah ofc, first it will check internally whether the particular product that they want to add exists in the db or not. If that exists already, a list of those available products must be displayed, then they must see an option to edit the existing product if they want to put some additional data if required into it.

Bhavishya2601: you are not getting it ig

see right now, user can check for a product in search bar by using barcode, if the product is found then he gets redirected to that specific product's page (he can make changes in that existing product by edit option) and if the product is not found then he gets the option to add it

Now from what I understood that you are trying to say is
Add a button on homepage to add product, which will follow the same approach as search functionality, first check if the product is found, if found redirects the user and if not, then let him add the product

Then why so we need this approach and also it will result in repetition of approach and code

swastikCommits: Yeah you are right though, but the important thing is adding a new product seems to be a quite important feature that the user might not know that it already exists as well when a "product is not found". 
Hence putting it on the homepage or somewhere else clearly visible can be a good idea so that even a new user can have eye on it and add a new product if he/she wants to.

Bhavishya2601: I don't think the button can be placed on the homepage as it doesn't make sense based on the functionality we have implemented as the user wouldn't be able to type the barcode, the barcode gets passed from the search page to avoid duplicacy

> the important thing is adding a new product seems to be a quite important feature that the user might not know that it already exists as well when a "product is not found".

we can think about this but I am not in favor of adding a btn on homepage.
wdyt @VaiTon ??

swastikCommits: I just wanted to put this idea on the table, that how new user can also have a better experience. 
Not precisely tbh a button on the homepage. It might a beautiful div as well or an option in the footer/navbar too? We can discuss it tho.

src/lib/api/product.ts
```diff
@@ -220,7 +220,8 @@ export type SelectedImage = {
 	y2: string;
 };
 
-type RawImage = {
+export type RawImage = {
+	url: '';
```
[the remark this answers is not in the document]
Bhavishya2601: My bad

Bhavishya2601:
@VaiTon Can you review this PR, and let me know if you need any changes

Bhavishya2601:
>Bhavishya2601 removed their assignment

Sry, I removed my assignment by mistake

Bhavishya2601:
@VaiTon  Can you review this PR, and let me know if you need any changes

src/routes/products/[barcode]/edit/+page.ts
```diff
@@ -14,7 +14,21 @@ export const load = (async ({ fetch, params }) => {
 		getTaxo<Country>('countries', fetch)
 	]);
 
-	if (product.status === 'failure') {
+	if (product.status === 'failure' && product.result.id === 'product_not_found') {
+		return {
+			state: {
+				status: 'empty',
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/+page.ts:20]: Extract this to a constant, so that we can compare to it instead of using `'empty'` one time in +page.ts and another in +page.svelte

VaiTon (approved):

Bhavishya2601:
@VaiTon Check out now

Issue #331, opened 2025-03-22
chore: Make Base URL Editable via `.env`


swastikCommits:
### Problem
The base URL (e.g., for API calls or redirects) may not be editable via a `.env `file, requiring hardcoding or redeployment for changes, especially in production on Vercel.
@VaiTon   in PR #243  suggested making it editable.

**This would improve flexibility, especially for Vercel production.**

### Proposed solution
Set base URL via `BASE_URL` in `.env`, with a fallback if unset.

PR #345, opened 2025-03-23
chore: made the base URL configurable via environment variables

swastikCommits:
### What
- Added `VITE_BASE_URL` to `.env.example` with default value `https://world.openfoodfacts.org`
- Updated `src/lib/const.ts` to use this environment variable with a fallback
- Replaced hardcoded URLs in` src/routes/products/search/+page.ts` and `src/lib/knowledgepanels/Action.svelte` to use the API_HOST constant

### Fixes bug(s)
#331

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,2 @@
 VITE_FOLKSONOMY_API_URL=https://api.folksonomy.openfoodfacts.net
-VITE_BASE_URL=https://world.openfoodfacts.org
+VITE_OFF_BASE_URL=https://world.openfoodfacts.org
--- a/.env.example
+++ b/.env.example
@@ -1,1 +1,2 @@
 VITE_FOLKSONOMY_API_URL=https://api.folksonomy.openfoodfacts.net
+VITE_BASE_URL=https://world.openfoodfacts.org
--- a/src/lib/knowledgepanels/Action.svelte
+++ b/src/lib/knowledgepanels/Action.svelte
@@ -1,10 +1,11 @@
 <script lang="ts">
 	import type { KnowledgeActionElement, KnowledgePanel } from '$lib/api';
 	import { goto } from '$app/navigation';
+	import { API_HOST } from '$lib/const';
 
 	// URL constants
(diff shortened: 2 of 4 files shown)


src/lib/knowledgepanels/Action.svelte
```diff
@@ -1,10 +1,11 @@
 <script lang="ts">
 	import type { KnowledgeActionElement, KnowledgePanel } from '$lib/api';
 	import { goto } from '$app/navigation';
+	import { API_HOST } from '$lib/const';
 
 	// URL constants
-	const PRODUCT_EDIT_URL = 'https://world.openfoodfacts.org/product/';
-	const PRODUCT_REPORT_URL = 'https://world.openfoodfacts.org/product/';
+	const PRODUCT_EDIT_URL = `${API_HOST}/product/`;
```
VaiTon, line 7: While we're here, can we move this to the `const.ts` file?

VaiTon (commented):
Just a little change and we're good

src/lib/knowledgepanels/Action.svelte
swastikCommits: Yup, sure.
Moving both `PRODUCT_EDIT_URL` and `PRODUCT_EDIT_URL` to `const.ts`

.env.example
```diff
@@ -1 +1,2 @@
-VITE_FOLKSONOMY_API_URL=https://api.folksonomy.openfoodfacts.net
\ No newline at end of file
+VITE_FOLKSONOMY_API_URL=https://api.folksonomy.openfoodfacts.net
+VITE_BASE_URL=https://world.openfoodfacts.org
```
VaiTon, line 2: Sorry for being pedantic, but maybe it's better to prefix this with `OFF_`, just to be sure we're not overriding something used by SvelteKit

swastikCommits: Ok adding the prefix

VaiTon (approved):

VaiTon:
Thank you! :tada:

Issue #333, opened 2025-03-22
Missing VITE_PRICES_API_URL in .env.example


krishna-shrivastav232702:
While exploring the codebase of openfoodfacts-explorer, i noticed that VITE_PRICES_API_URL is missing , and this url acts as base url to create a client , which provides methods for making http requests like get,post,put .....

You can find the reference to VITE_PRICES_API_URL in src/lib/api/prices.ts

based on my understanding , the intended url might be https://prices.openfoodfacts.org , but i am not sure

PR #346, opened 2025-03-23
chore: added the missing VITE_PRICES_API_URL

swastikCommits:
### What
- Added `VITE_PRICES_API_URL=https://prices.openfoodfacts.org` to `.env.example`
- This ensures the prices API client in `src/lib/api/prices.ts` has its base URL properly configured

### Fixes bug(s)
#333


VaiTon:
That env var should be not set by default. That's why it has a conditional in the API .ts file. That's because setting the var is a flag to enable the feature. Also, I would prefer if no .example vars are pointing to prod websites, as devs will `mv .env.examples .env` and then forget about it.

PR #349, opened 2025-03-23
style: adjust layout and spacing in Nova and NutriScore components

Abbas-Askari:
### What
Fix the overflow show below

### Screenshot
Before:
image
After:
image


Abbas-Askari:
> I'm sorry @Abbas-Askari, I think I already fixed it in a commit. Can you confirm?

Yes, It is fixed. Closing.

PR #350, opened 2025-03-23
feat: Added Browser Caching for Products with Questions on HomePage

heymitali:
### What
- Optimized home page performance by implementing caching for product listings, reducing redundant API calls. This enhancement significantly boosts site responsiveness and reduce backend server load.
- This is a significant performance improvement for the home page. Initially, it made 11 API calls altogether—first fetching products using the Questions API, and then making a Products API call for each of the 10 products. All of these calls will be eliminated with the use of caching.

## Changes
- Implemented browser-side caching using `svelte-local-storage-store` for products with questions data
- Cache expires after 10 minutes to ensure data freshness
- Added type safety with `CachedData` interface

## Performance Impact
- Reduces API calls to OpenFoodFacts API by caching results in localStorage
- Improves subsequent page load times within cache window (10 minutes)
- Eliminates redundant API calls for the same data
- Maintains data freshness with automatic cache invalidation

## Technical Details
- Cache key: `cached-products-with-questions`
- Cache structure includes both products data and timestamp
- Cache expiry: 10 minutes
- Graceful fallback to API calls when cache is invalid or expired

### Screenshot
All API hits related to product list:

Screenshot 2025-03-24 at 3 11 45 AM

Demo Recording:

https://github.com/user-attachments/assets/99f05968-e375-429c-b7a7-5074e453fcd3

### DIFF
--- a/src/routes/+page.ts
+++ b/src/routes/+page.ts
@@ -5,6 +5,8 @@ import {
 	type ProductStateFound
 } from '$lib/api';
 import type { PageLoad } from './$types';
+import { persisted } from 'svelte-local-storage-store';
+import { get } from 'svelte/store';
 
 type QuestionsResponse = {
 	status: string;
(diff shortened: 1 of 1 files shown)


VaiTon:
Hey @heymitali, thank you for your PR. 

Please next time discuss issues and how you intend to implement changes before doing that, otherwise I can't assure that I will be able to merge the PR.

The problem I see with this PR is that I don't really want to cache anything here apart from the first request to the robotoff API (because it's too slow. I would not even cache that one).

Images should be already cached by the browser (or that's another issue), and fetching the jsons could be done more effectively (maybe with just a request?).

heymitali:
Sure, i will keep this in mind now onwards. Thanks.

PR #351, opened 2025-03-24
feat: Added configurable radius parameter for searching nearby stores

heymitali:
### What
- In this PR, I did couple of improvements in location library file.
- I have added necessary parameters for improving accuracy and better results while fetching current position. I have used these parameters: https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition#options
- function getNearStores() was internally hardcoded to use 1000 as fixed radius always, which is not good. The function should take radius as input parameter and should search accordingly. I have implemented this change in this function, along with improving other aspects of code.


heymitali:
@teolemon In this PR, I did changes only in location lib file and not in any frontend code. We can do the frontend changes later on in separate PR. Since now the radius parameter is configurable, it should be doable.

src/lib/location.ts
```diff
@@ -21,23 +26,38 @@ export type OverpassAPIResult = {
...
+		if (!response.ok) {
+			throw new Error(`Overpass API error: ${response.status} ${response.statusText}`);
+		}
+
+		const data = await response.json();
+		return data as OverpassAPIResult;
+	} catch (error) {
+		console.error('Error fetching nearby stores:', error);
+		throw error;
+	}
```
VaiTon, line 60: This block is useless, I would remove the try/catch entirely

heymitali: @VaiTon  I have removed this. Please review again.

VaiTon (approved):
Thank you!

Issue #287, opened 2025-03-14
The server is currently kind of down. Ensure we have a better behavior when the server is down.


teolemon:
### What
The server is currently kind of down. Not sure if that's the ground cause of the screenshot error. Ensure we have a better behavior when the server is down.
cc @VaiTon 

![Image](https://github.com/user-attachments/assets/1df71ede-e3b4-48bc-ac44-36a2e5eeec7d)

lakshaydewan:
@teolemon  would love to work on this.

KarruHarin:
@teolemon Is there  way to reproduce the error or server crash , I am unable to recreate it just based on the image .

VaiTon:
> [@teolemon](https://github.com/teolemon) Is there way to reproduce the error or server crash , I am unable to recreate it just based on the image .

Disable your internet access and try to open the dev server web page

github-throwaway:
@akashshetty1997 could you check if this is still relevant? Thanks.

akashshetty1997:
sure @github-throwaway  i will look into it.

akashshetty1997:
Currently, when you disable your internet access and try to open the dev server webpage, this is the behavior you’ll see:

https://github.com/user-attachments/assets/be139526-3ac7-409e-94d2-20e510ffbb8a

Do you think we should show a user-friendly error message like the example below?

![Image](https://github.com/user-attachments/assets/243c2f95-bdc9-4b46-8941-cb04c404ef99)

@VaiTon  @github-throwaway

github-throwaway:
Your video does not play for me. Could post a screenshot?

akashshetty1997:
1. If the internet is off and I try to run the project, the server closes automatically and throws the following error:

![Image](https://github.com/user-attachments/assets/99aa7c2d-f54e-43ed-9908-8c312e208364)

2. However, if I keep the internet on, let the project fully load, and then turn off the internet, the following behavior is observed instead:

![Image](https://github.com/user-attachments/assets/c619178b-f5b4-44bd-b6a5-c4ac88058d4a)

 @github-throwaway

akashshetty1997:
@github-throwaway  What’s the best way for us to handle backend/API unavailability on the frontend in a clean, user-friendly way? Should we go with a full error page or just show fallback components? Also wondering how we can make sure we're covering things like try/catch around API calls, global error handling, retry logic, and offline detection.

github-throwaway:
[...]

PR #364, opened 2025-03-27
fix: Undefined issue was causing blank screen when launched

pascalypperciel:
### What
When `prices.results` is undefined, it causes `.length` to throw an error. The error seen in the provided screenshot was: `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'length')`, and after looking into the stack trace myself, I saw that the error pointed to this line in the component:
Screenshot_4

I was then able to reproduce the bug locally:
Screenshot_2

I fixed the bug by using optional chaining to safely access nested properties. This prevents the app from crashing when `prices.results` is undefined and avoids calling `.length` on undefined like before.

### Screenshot
With my change, the blank page behaviour wasn't recurring.
Screenshot_3

### Fixes bug(s)
#287

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,10 @@
+{
+	"name": "OpenFoodFacts Explorer",
+	"customizations": {
+		"vscode": {
+			"extensions": ["svelte.svelte-vscode"]
+		}
+	},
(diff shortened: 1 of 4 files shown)


VaiTon (approved):

VaiTon:
Hey, thanks for the PR! Please rebase on main / fix the svelte-check issue =)

PR #369, opened 2025-03-28
feat: Add product comparison feature

Sunhao86:
### What
This PR adds a product comparison feature to allow users to compare multiple food products side by side

### Screenshot
截图 2025-03-27 20-16-48
截图 2025-03-27 20-16-53

### Features
- Users can add products to a comparison list from product detail pages
- Users can view all selected products in a comparison view
- Users can remove products or clear the entire list

### DIFF
--- a/src/routes/compare/+page.svelte
+++ b/src/routes/compare/+page.svelte
@@ -4,17 +4,30 @@
 	import { compareStore } from '$lib/stores/compareStore';
 
 	function getNutriscoreImage(grade: string | null | undefined) {
-		return KP_ATTRIBUTE_IMG('nutriscore-' + grade + '-new-en.svg');
+		return KP_ATTRIBUTE_IMG('nutriscore-' + (grade ?? 'unknown') + '-new-en.svg');
 	}
 
-	function getNovaImage(group: string | number) {
+	function getNovaImage(group: string | number | null | undefined) {
 		return group
 			? KP_ATTRIBUTE_IMG('nova-group-' + group + '.svg')
 			: KP_ATTRIBUTE_IMG('nova-group-unknown.svg');
 	}
 
-	function getEcoscoreImage(grade: string) {
-		return KP_ATTRIBUTE_IMG('ecoscore-' + grade + '.svg');
+	function getEcoscoreImage(grade: string | null | undefined) {
+		return KP_ATTRIBUTE_IMG('ecoscore-' + (grade ?? 'unknown') + '.svg');
+	}
+
+	// Helper to format nutrient values
+	function formatNutrient(value: number | undefined | null, decimals = 1): string {
+		if (value === undefined || value === null) {
+			return '-';
+		}
+		// Show 0 decimals for energy
+		if (decimals === 0) {
+			return value.toFixed(0);
+		}
+		// Avoid showing .0 for whole numbers like 5.0g -> 5g
+		return String(parseFloat(value.toFixed(decimals)));
 	}
 </script>
 
@@ -23,7 +36,7 @@
 	<meta name="description" content="Compare nutritional information of food products" />
 </svelte:head>
 
-<Card>
+<Card class="max-w-6xl mx-auto px-4">
 	<div class="mb-4 flex items-center justify-between">
 		<h1 class="text-2xl font-bold">Compare Products</h1>
 		{#if $compareStore.length > 0}
@@ -40,22 +53,22 @@
 		</div>
 	{:else}
 		<div class="overflow-x-auto pb-4">
-			<div class="grid grid-cols-1 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4 gap-4">
+			<div class="flex flex-wrap justify-center gap-4">
 				{#each $compareStore as product (product.code)}
-					<div class="relative rounded-lg border p-2">
+					<div class="relative rounded-lg border p-2 shadow-md w-64 flex flex-col">
 						<button
-							class="btn btn-circle btn-xs btn-ghost absolute top-2 right-2"
+							class="absolute top-2 right-2 rounded-full bg-red-500 text-white p-1 hover:bg-red-600 cursor-pointer z-10"
 							onclick={() => compareStore.removeProduct(product.code)}
 							aria-label="Remove product from comparison"
 						>
-							<span class="icon-[mdi--close] h-3 w-3"></span>
+							<span class="icon-[mdi--close] h-4 w-4 block"></span>
 						</button>
 
-						<div class="mb-4 flex flex-col items-center pt-4">
+						<div class="flex flex-col items-center pt-4">
 							{#if product.image_front_small_url}
 								<img
 									src={product.image_front_small_url}
-									alt={product.product_name}
+									alt={product.product_name ?? product.code}
 									class="mb-2 h-24 object-contain"
 								/>
 							{/if}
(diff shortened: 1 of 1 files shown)


Sunhao86:
I've implemented this product comparison feature to help users compare food products side by side. I genuinely welcome any feedback or suggestions for improvement. Thank you for reviewing my PR 🙏

teolemon:
- Can you add the screenshot of the compare system as well ? @Sunhao86 
- One interesting thing would be infinite side by side comparison, but the product battle mode is cool already.

src/lib/stores/compareStore.ts
```diff
@@ -0,0 +1,51 @@
...
+			return saved ? JSON.parse(saved) : [];
+		}
+		return [];
+	};
+
+	const { subscribe, update, set } = writable<ProductReduced[]>(getInitialState());
+
+	if (browser) {
+		subscribe((products) => {
+			localStorage.setItem('compareProducts', JSON.stringify(products));
```
VaiTon, line 20: Instead of this please use `stored` with `svelte-local-storage-store`

VaiTon (changes requested):

Sunhao86:
Hi @VaiTon ,I've updated the compareStore to use `svelte-local-storage-store` as requested. Please kindly review again 🙏

Sunhao86:
Hi @teolemon ,Thank you for the feedback!  As for the screenshot, here’s how the current comparison system looks:

Currently, the comparison supports up to 4 products side by side. I’ve limited it to 4 because adding more makes the layout messy, especially on smaller screens. But I'm happy to explore infinite scroll-style layouts in the future!

Let me know if you'd like me to tweak anything further 🙏
image

src/lib/ui/CompareFloatingButton.svelte
```diff
@@ -0,0 +1,20 @@
+<script lang="ts">
+	import { compareStore } from '$lib/stores/compareStore';
+
+	let count = $state(0);
+
+	$effect(() => {
+		count = $compareStore.length;
```
VaiTon, line 7: This should be a $derived

src/routes/compare/+page.svelte
```diff
@@ -0,0 +1,114 @@
...
+			<p class="mb-4 text-lg">No products selected for comparison</p>
+			<a href="/products/search?q=chocolate" class="btn btn-primary">Browse Products</a>
+		</div>
+	{:else}
+		<div class={isMobile ? 'overflow-x-auto pb-4' : ''}>
+			<div
+				class={isMobile ? 'w-max' : ''}
+				style={isMobile
+					? ''
+					: `display: grid; grid-template-columns: repeat(${$compareStore.length}, 1fr); gap: 1rem;`}
```
VaiTon, line 59: Please use TailwindCSS instead of custom styles. Also, is the isMobile var really necessary? Can't we do that in CSS?

src/routes/compare/+page.ts
```diff
@@ -0,0 +1,8 @@
+import type { PageLoad } from './$types';
+
+export const load: PageLoad = async () => {
+	return {
+		title: 'Compare Products',
+		description: 'Compare nutritional information of food products'
```
VaiTon, line 6: What is the utility of this file? Couldn't we just embed the text inside the .svelte file?

VaiTon (changes requested):

teolemon:
What I'd like to see as screenshot is how you add a product to the comparison @Sunhao86

Sunhao86:
hi @teolemon , here are the screenshots showing how a product is added to the comparison:

From the product detail page, I click on "Add to Compare"

Then, I go to the Compare Products page to see all selected products displayed side by side.
截图 2025-04-04 16-07-29
截图 2025-04-04 16-09-15

Sunhao86:
Hi @VaiTon , I’ve implemented all your suggestions:

replaced $state with $derived

removed isMobile and switched to TailwindCSS

deleted the unused +page.ts

Everything works well locally. Please feel free to share more feedback — I’ll be happy to adjust anything needed. Thanks a lot!

VaiTon:
@github-throwaway @teolemon UX feedback?

github-throwaway:
Nice idea!

- It looks like the X symbol does not play nice with large images. I would maybe move that and make it more prominent
- I'm unsure about the rectangle images and the rounded bounding boxes. I feel that does not match the style we have on the landing page. But I also don't have a concrete solution to this. I just think it stands out.
- the drop shadow is missing if we wanna stay consistent with the landing page. 
- unrelated but we really need to fix the margins in the navbar. They are all over the place... 
- it terms of comparability it's a good start. But I would expect to see nutrition facts and ingredients side by side. The current state is a bit superficial (a great start though, well done!)
- what happens when I click clear all?
- what does this look like on mobile? (Product pages didn't load for me on iOS chrome)

cc: @abrahamvg

abrahamvg:
Suggestions:
- The container should have a fixed width otherwise it will expand and not look good when there are less than 4 items.
- The X can be wrapped inside a circle container with the bg being red
- `Hot take` : instead of the images for nutritional scores, can be add it as a value ? or we can stack it vertically to increase readability

github-throwaway:
Also what happens when I click the compare button in the lower right corner?

Bhavishya2601:
we already have 4 buttons there, adding a 5th btn there (Add to compare), I think so would affect the UI

In this product if you see, It's already quite suffocated
image


@Sunhao86 can you pls provide a screenshot of the product page where the btn (Add to compare) will be mounted after pulling all the new commits

swastikCommits:
We can replace the text inside these buttons by replacing it with the following logos:
Add to calculator: https://icon-sets.iconify.design/ic/?icon-filter=calc&keyword=p
Add to compare: https://icon-sets.iconify.design/ic/?icon-filter=compa&keyword=p
Edit: https://icon-sets.iconify.design/ic/?icon-filter=edit&keyword=p
Share: Just by removing the text inside the button

These buttons on hover must specify "Add to calculator". "Add to compare", "Edit", "Share"
wby @github-throwaway and @VaiTon ?

github-throwaway:
@Sunhao86 Please update your feature branch to include the latest changes of the main branch. 

On mobile we should use icons. And on desktop text + icons.

Sunhao86:
> @Sunhao86 Please update your feature branch to include the latest changes of the main branch.
> 
> On mobile we should use icons. And on desktop text + icons.

Hi @github-throwaway , I've updated the feature branch with the latest changes from `main` and resolved the merge conflicts, as requested.

Sunhao86:
Hi @VaiTon  @github-throwaway  ,  Thanks for the great suggestions!
We've updated the CompareFloatingButton to avoid overlapping with the calculator button on mobile view.
It now appears on the left bottom corner, while the calculator stays on the right. I'm really thank for your UI/UX suggestions!
Screenshot 2025-04-13 170115
Screenshot 2025-04-13 170100

Bhavishya2601:
@Sunhao86 Why do we even need a floating button for compare? can't we place it next to folksonomy as it is redirecting us to compare page and is not a pop up like nutrition calculator

and pls fix this
image


@VaiTon @github-throwaway don't you think compare feature and nutrition calculator are quite same in some perspectives
image

Sunhao86:
@Bhavishya2601 Thanks for the feedback! Thank you very much for your advice!

I understand the concern about having multiple floating buttons, especially on mobile where screen space is limited.

As an alternative, I was wondering if we could move the compare functionality to a more dedicated place — for example, a separate page 
This would avoid floating buttons altogether while still preserving quick access to comparisons.

What do you think about this direction? @github-throwaway

github-throwaway:
It might make sense to combine these things. They are indeed quite comparable.

 The dedicated page does not make sense to me. I think most features should be available from the product page directly.

VaiTon:
> @VaiTon @github-throwaway don't you think compare feature and nutrition calculator are quite same in some perspectives

Not necessarily, but if we find a way to integrate them I'm for it

github-throwaway:
please provide a mockup of your alternative compare page. You are allowed to dislike something but that’s not constructive criticism. Instead address concrete pain points. 

The nutritional info is not easily comparable. The info should be on the same row. Maybe some sort of table view would make more sense here?

Sunhao86:
@github-throwaway  @VaiTon @Bhavishya2601  Thank you all so much for your valuable comments and suggestions recently!
@github-throwaway Thank you for further elaborating on the idea of ​​a dedicated compare page and your preference for integrating the functionality directly into the product page. However, in order to better realize your vision, I would like to ask:
1. Regarding "providing functionality directly from the product page", what specific interaction method do you prefer? For example, is it to pop up a dialog box (modal) to display the comparison information after clicking a button on the product page? Or dynamically embed a comparison area in the current page? Or are there other ways or reference examples that you think are better for the experience that you can share?
2. You mentioned using a table view (rows are nutrients, columns are products) to compare nutritional information. This idea sounds great and is indeed more conducive to direct comparison. Is this table view also used in your envisioned product page integration solution?
@VaiTon  @Bhavishya2601 Thank you for your thoughts on integrating the comparison and calculator functions and your feedback on the current comparison page UI. If we determine a new design for product page integration, we can discuss how to better combine these ideas.

Thanks again for everyone's help!

teolemon:
@VaiTon on mobile yes
Lists - Light - En
Side by side Compare - Light - EN-1
Side by side Compare - Light - EN

PR #371, opened 2025-03-28
fix: ...Undefined issue was causing blank screen when launched

pascalypperciel:
### What
When `prices.results` is undefined, it causes `.length` to throw an error. The error seen in the provided screenshot was: `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'length')`, and after looking into the stack trace myself, I saw that the error pointed to this line in the component:
Screenshot_4

I was then able to reproduce the bug locally:
Screenshot_2

I fixed the bug by using optional chaining to safely access nested properties. This prevents the app from crashing when `prices.results` is undefined and avoids calling `.length` on undefined like before.

### Screenshot
With my change, the blank page behaviour wasn't recurring.
Screenshot_3

### Fixes bug(s)
#287

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -11,3 +11,4 @@ node_modules
 pnpm-lock.yaml
 package-lock.json
 yarn.lock
+.pnpm-store/
--- a/src/routes/products/[barcode]/Prices.svelte
+++ b/src/routes/products/[barcode]/Prices.svelte
@@ -144,7 +144,7 @@
 <div>
 	<div id="prices">
 		<span class="font-bold">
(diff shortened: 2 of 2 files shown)


pascalypperciel:
Hello @VaiTon,

I am a bit embarassed, I messed up with my `git` because I got overconfident and used the CLI. I decided it was probably cleaner to just make a new PR. Sorry!

I added the line in `.prettierignore` because when I ran the workflows with `act` I would get errors. It's a common folder to put in this file so I humbly think that it's alright.

VaiTon (approved):

PR #376, opened 2025-03-28
ci: auto-generate-changelogs

teolemon:
### What
- I propose we do time based releases, just for the sake of it, since we already deploy at every commit

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -1,0 +1,28 @@
+name: 🚀 Adding the commit to the next release
+on:
+  push:
+    branches:
+      - main
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - id: release
+        uses: googleapis/release-please-action@v4.1.4
+        with:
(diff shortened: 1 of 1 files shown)


VaiTon:
@teolemon do we need semantic versioning for a project like this? I mean, we could have a "stable" branch and regularly merge from main / devel. 

We don't need to add complexity with semantic versioning as we have no users (in terms of API).

teolemon:
no semver needed, it's more regularly release. The only goal is to generate changelogs, create excitement, and post the novelties in slack @VaiTon

github-throwaway:
Does this really make sense? Feels like one of these things that we have on the main website/repo so we are trying to clone it to this without having an actual need for it.

VaiTon:
> The only goal is to generate changelogs, create excitement, and post the novelties in slack @VaiTon

Oh that's right with me, but I don't like GitHub releases with semver versioning. Is there any way to change that?

github-throwaway:
Related discussion:

https://github.com/openfoodfacts/openfoodfacts-explorer/discussions/385

PR #386, opened 2025-03-28
style: X logo color changes as per the theme

swastikCommits:
### What
Fixes the color issue of X logo in case of light mode.

### Screenshot
Before:
Screenshot 2025-03-28 222344

After: 
Screenshot 2025-03-28 222334


swastikCommits:
Removed the `@iconify/logos` dependency which was used earlier by the X logo

Bhavishya2601:
@swastikCommits I don't think there's any point on fixing the X's logo theme as we are already discussing to change the footer

swastikCommits:
@Bhavishya2601 the footer design would still have a X logo right? Nevertheless, it would still work as a temporary fix until the footer is completely redesigned

Bhavishya2601:
@swastikCommits Bro it's upto the mentors, but I don't think working on this is meaningful, as it will redesigned soon

swastikCommits:
No worries, as the discussion about footer design is still going on and the design haven't been finalized yet.

@VaiTon what's your opinion? I think until and unless the footer has been redesigned this would still be beneficial for the users even if this design stays for a couple of days

Bhavishya2601:
@swastikCommits @VaiTon I was also saying to not work on this because there already a PR open on footer for adding badges, making changes in the footer file might end up in merge conflicts

swastikCommits:
@Bhavishya2601 I looked over it too, so ig it only adds badges to the footer(which is an addition to the footer). Does it make changes to logos too?  If I'm wrong please correct me.
If it doesn't make any changes to the logos part, then there won't be any merge conflicts.

Bhavishya2601:
yeah you are right, It is addition to the footer, I made comment without looking through your code, so my apologese 😅

>I don't think there's any point on fixing the X's logo theme as we are already discussing to change the footer

But the first reason is still valid in my view😅

swastikCommits:
No worries @Bhavishya2601 
Let the maintainers choose if we should keep it or not. As the footer designs haven't been finalized yet, and this would create a impact on UX with a quick merge(coz of its simplicity)

VaiTon (approved):

VaiTon:
@swastikCommits please rebase on main

swastikCommits:
I have made the changes 👍

Issue #361, opened 2025-03-26
Various issues when trying to load a product


teolemon:
### What
- Various issues when trying to load a product: https://openfoodfacts-explorer.vercel.app/products/3100881040809
- Missing translations (we probably need to validate the Crowdin PR)
- Robotoff CORS error
- Missing property

Also it complains of an hydration mismatch on a bottle of water 🙃 

### Steps to reproduce the behavior:

![Image](https://github.com/user-attachments/assets/fc011d2a-55f9-40af-9c0c-949e83199dd9)

swastikCommits:
@teolemon I would like to work on it

swastikCommits:
The fix in #387  addresses this issue since I could only see the translations and a `bgColor` error in my console. I believe the type error and CORS error might have been fixed in a different commit. If I am mistaken, please let me know.

PR #387, opened 2025-03-29
fix: type errors fixed in a product which fails to fetch

swastikCommits:
### What
- We currently use a grading system which furthers used to map, this console error of `bgColor` arises due to grade being undefined while destructring.
So using a fallback of `( || 'unknown')` for grade whenever it is undefined/null fixes the issue

- Used Crowdin to fix the translations in #406 

### Screenshot
Before: blank Screen, as it could not fetch the data from the BE

After: 
image

### Fixes bug(s)
#361

### DIFF
--- a/src/routes/products/[barcode]/+page.svelte
+++ b/src/routes/products/[barcode]/+page.svelte
@@ -181,7 +181,7 @@
 		<Nova grade={product.nova_group} />
 	</a>
 	<a href="#environment_card" class="md:w-1/3">
-		<EcoScore grade={product.ecoscore_grade || 'unknown'} />
+		<EcoScore grade={product.ecoscore_grade ?? 'unknown'} />
 	</a>
 </div>
 
--- a/src/routes/products/[barcode]/+page.svelte
+++ b/src/routes/products/[barcode]/+page.svelte
@@ -181,7 +181,7 @@
 		<Nova grade={product.nova_group} />
 	</a>
 	<a href="#environment_card" class="md:w-1/3">
-		<EcoScore grade={product.ecoscore_grade} />
+		<EcoScore grade={product.ecoscore_grade || 'unknown'} />
 	</a>
 </div>
 
--- a/src/routes/products/[barcode]/GreenScore.svelte
+++ b/src/routes/products/[barcode]/GreenScore.svelte
@@ -40,7 +40,7 @@
 	let { grade }: { grade: string } = $props();
 	let normalizedGrade = $derived(grade.toLowerCase());
 	let { name, src, textColor, bgColor } = $derived(
-		GRADE_MAP[normalizedGrade] || GRADE_MAP['unknown']
+		GRADE_MAP[normalizedGrade] ?? GRADE_MAP['unknown']
 	);
 </script>
 
--- a/src/routes/products/[barcode]/GreenScore.svelte
+++ b/src/routes/products/[barcode]/GreenScore.svelte
@@ -39,7 +39,9 @@
 
 	let { grade = 'unknown' }: { grade?: string } = $props();
 	let normalizedGrade = $derived((grade || '').toLowerCase());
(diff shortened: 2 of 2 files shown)


src/routes/products/[barcode]/GreenScore.svelte
```diff
@@ -37,9 +37,11 @@
 		}
 	};
 
-	let { grade }: { grade: string } = $props();
-	let normalizedGrade = $derived(grade.toLowerCase());
-	let { name, src, textColor, bgColor } = $derived(GRADE_MAP[normalizedGrade]);
+	let { grade = 'unknown' }: { grade?: string } = $props();
```
Comment (VaiTon) [inline src/routes/products/[barcode]/GreenScore.svelte:40]: `grade` here should not be an optional prop.

VaiTon (changes requested):

src/routes/products/[barcode]/GreenScore.svelte
Comment (VaiTon) [reply src/routes/products/[barcode]/GreenScore.svelte:40]: The issue should not be fixed here, but at the calling site

Comment (swastikCommits) [reply src/routes/products/[barcode]/GreenScore.svelte:40]: Added the OR operator( `|| 'unknown'`) in the EcoScore component

Comment (sohan-gupthak) [reply src/routes/products/[barcode]/GreenScore.svelte]: hey @swastikCommits, 
Instead of an empty string, you might consider defaulting to 'unknown' directly
`let normalizedGrade = $derived((grade || 'unknown').toLowerCase());`

Comment (swastikCommits) [reply src/routes/products/[barcode]/GreenScore.svelte]: @sohan-gupthak as requested by @VaiTon `'unknown'`was assigned to the calling page in EcoScore component.

Thanks for the suggestion about adding a console log alert for missing grade. @VaiTon wby should this be added as a check in console?

Comment (swastikCommits) [reply src/routes/products/[barcode]/GreenScore.svelte]: I have removed the null check from `grade` as it has already been handled by the calling site.

src/routes/products/[barcode]/GreenScore.svelte
```diff
@@ -39,7 +39,9 @@
 
 	let { grade }: { grade: string } = $props();
 	let normalizedGrade = $derived(grade.toLowerCase());
-	let { name, src, textColor, bgColor } = $derived(GRADE_MAP[normalizedGrade]);
+	let { name, src, textColor, bgColor } = $derived(
+		GRADE_MAP[normalizedGrade] || GRADE_MAP['unknown']
+	);
 </script>
```
Comment (VaiTon) [inline src/routes/products/[barcode]/GreenScore.svelte:45]: Let's check in a single place only.

VaiTon (changes requested):

src/routes/products/[barcode]/+page.svelte
```diff
@@ -181,7 +181,7 @@
 		<Nova grade={product.nova_group} />
 	</a>
 	<a href="#environment_card" class="md:w-1/3">
-		<EcoScore grade={product.ecoscore_grade} />
+		<EcoScore grade={product.ecoscore_grade || 'unknown'} />
```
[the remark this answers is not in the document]
Comment (swastikCommits) [reply src/routes/products/[barcode]/+page.svelte:184]: Yeah, got it. I'll keep this in mind.

src/routes/products/[barcode]/GreenScore.svelte
Comment (swastikCommits) [reply src/routes/products/[barcode]/GreenScore.svelte:45]: Yeah, my bad I forgot to remove this check here.

Comment (swastikCommits) [reply src/routes/products/[barcode]/GreenScore.svelte:45]: After I removed the check here, it broke again, not fetching the data.
After debugging, I found out that removing the check `GRADE_MAP['unknown']` isn't valid coz always it isn't always sure that the parent component will provide a valid grade.
Even with the parent check, the component could receive values(lowercase) that aren't in the `GRADE_MAP` for example something else like 'k' instead of valid grades - 'a', 'b', 'c', 'd'

VaiTon (approved):

Issue #377, opened 2025-03-28
Make Open Food Facts Explorer fully translatable


teolemon:
### What

- [ ] Make Open Food Facts Explorer fully translatable
- [ ] Honor the language selection in settings
- [ ] Make all strings translatable
- [ ] Set up translation using Crowdin

Bhavishya2601:
I would like to work on this issue

PR #388, opened 2025-03-29
feat: made all strings translatable using Crowdin

swastikCommits:
### What
- Made all the strings translatable.
- Added the translations to both "en" and "it"
- Updated the components that uses the translation  

### Part of 
#377


github-throwaway:
Please use Crowdin (as clearly stated in the issue)

https://crowdin.com/blog/2024/03/12/svelte-localization

swastikCommits:
Thankyou @github-throwaway I will soon add the changes

github-throwaway:
Also, I think the prefixed chore is incorrect here. It should be feat. 

https://stackoverflow.com/a/26944812

swastikCommits:
Thankyou will keep in mind to change it to feat after I push the commit with the changes

swastikCommits:
@github-throwaway have a look now, installed svelte-i8n and used the Cowdin CLI for the translations.
Just review `crowdin.yml` too, what shall I use as the 'project_id' and 'api_token'

github-throwaway:
@teolemon @VaiTon do we need a maintainer to set up the repo to work with crowdin?

swastikCommits:
Actually we need to add `crowdin extract`, `crowdin upload sources` and `crowdin download` as scripts to `package.json`

package.json
```diff
@@ -70,5 +71,8 @@
 			"@vercel/speed-insights",
 			"esbuild"
 		]
+	},
+	"dependencies": {
+		"svelte-i18n": "^4.0.1"
```
VaiTon, line 76: I think we can move this to dev-dependencies

VaiTon (changes requested):
Please rebase on main

package.json
swastikCommits: Moved it to dev-dependencies 👍

swastikCommits:
@VaiTon I'm getting issues due to patch, I may close this PR and make a fresh one

Issue #403, opened 2025-03-31
Add a way to specify API endpoint credentials


github-throwaway:
> I think there is something wrong with api, I get cors error when I try to go to products page locally,

> Can you check if it's me only or everyone is getting this error?

[Slack Message](https://openfoodfacts.slack.com/archives/C08FMJWP6J0/p1743171053107079?thread_ts=1743171053.107079&cid=C08FMJWP6J0)

~Removal of .env made the error go away. Possibly related to the auth required in the staging environment.~

Edit (by VaiTon): The staging endpoints (provided by default with the .env.example file) require basic authentication. We should:
- add two variables (username and password) to the env file to handle this
- add the required logic in the API layer to forward basic credentials to the API (if they are set)

Blitz-Cloud:
I found this on the docs regarding the auth part https://openfoodfacts.github.io/openfoodfacts-server/api/#authentication . As I understood we should change the user agent for the library that makes the requests if the staging server it is used

VaiTon:
Implemented in #729

PR #405, opened 2025-03-31
Update to env.example

Blitz-Cloud:
I updated the domain VITE_FOLKSONOMY_API_URL to point to https://api.folksonomy.openfoodfacts.org instead of https://api.folksonomy.openfoodfacts.net This should solve all the cors issues that other had with running the project

### What
I updated the domain VITE_FOLKSONOMY_API_URL to point to https://api.folksonomy.openfoodfacts.org instead of https://api.folksonomy.openfoodfacts.net This should solve all the cors issues that other had with running the project.

### Screenshot

image

### Fixes bug(s)
#403

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,2 @@
-VITE_FOLKSONOMY_API_URL=https://api.folksonomy.openfoodfacts.net
+VITE_FOLKSONOMY_API_URL=https://api.folksonomy.openfoodfacts.org
 VITE_OFF_BASE_URL=https://world.openfoodfacts.org


github-throwaway:
That's just updating the environment to the production environment. I'm not sure if that the right way to go here.
@VaiTon do you have a preference here?

Blitz-Cloud:
Is there any way to use the staging environment, because right now it is required to use an username and a password. If the solution from above isn't ok I will look into and also update the readme on how to get the username and password and also mention this in the readme for the project.

VaiTon:
Nop, I'm sorry but this is not the right way to solve the issue. To solve the CORS issue we should raise an issue against the relevant repository to fix it for the entire community. To solve the missing credentials we should solve #403

PR #407, opened 2025-04-01
fix: overlapping tags in edit page

swastikCommits:
### Fixes bug(s)
- #192

### DIFF
--- a/src/routes/products/[barcode]/edit/Tags.svelte
+++ b/src/routes/products/[barcode]/edit/Tags.svelte
@@ -46,7 +46,7 @@
 </script>
 
 <div
-	class="bg-base-100 flex h-auto min-h-12 w-full flex-wrap gap-x-1.5 gap-y-1 rounded-md p-2 border border-base-200 focus-within:border-primary focus-within:outline-2 focus-within:outline-primary focus-within:outline-offset-2"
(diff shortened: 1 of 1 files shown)


VaiTon:
> Added min-w-[100px] ensures to provide the container a minimum width, which prevents the tags from compressing

Please do not use absolute values unless necessary. Prefer tailwind `-md`, `-lg`, ...

swastikCommits:
The `flex` class was missing from the main container div that holds the tags, which causes the overlap.
It used `flex-wrap` which is a flex property, but the container wasn't a flex container in the first place itself.

src/routes/products/[barcode]/edit/Tags.svelte
```diff
@@ -46,23 +46,23 @@
...
+	class="input input-bordered bg-base-100 flex h-auto min-h-12 w-full flex-wrap gap-x-1.5 gap-y-1 rounded-md p-2"
 >
 	{#each tags as tag (tag)}
 		<span class="badge badge-ghost overflow-hidden py-3" transition:fade={{ duration: 100 }}>
 			<span class="truncate">{tag}</span>
 			<button class="ml-1 text-xl" onclick={removeTag(tag)}>×</button>
 		</span>
 	{/each}
-	<div class="dropdown grow">
+	<div class="dropdown min-w-[100px] grow">
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/Tags.svelte:57]: Do not use absolute values....

VaiTon (changes requested):

swastikCommits:
@VaiTon @github-throwaway I have pushed all the changes and confirmed its responsiveness for with the light/dark theme too.
These new classes completely replace the `input-bordered` class, which caused the user interface to break.
Screenshot 2025-04-04 064021
Screenshot 2025-04-09 055215

Issue #337, opened 2025-03-23
Product page (tracker)


teolemon:
### Product page
- [ ] Add support for Price visualisation on product page
- [ ] the knowledge panel action button do not work
- [ ] NOVA is distorted + wrong colors

Bhavishya2601:
>Add support for Price visualisation on product page

By this do you mean, we need to show the price of the product?

>NOVA is distorted + wrong colors

What's the issue with NOVA? Can you pls describe it in a little detail

Bhavishya2601:
>Add a "edit history" section like on the current website

I'm currently working on this issue

swastikCommits:
> * Add a "edit history" section like on the current website

As I deep dived into the main website's docs, these history fields like `created_t`, `last_modified_t`, `creator`, `created_datetime` and `last_modified_datetime` can be fetched from ProductApi's product object.
The snippet would look like this:
```
   const productApi = new ProductsApi(fetch);
   const product = await productApi.getProduct(barcode);
```
@Bhavishya2601 I hope it helps.

akashshetty1997:
Hey! I'd love to work on adding the sticky "Go up" button on the product page. 
Just wondering, should the design match the one on the current (old) website, or are we open to tweaking it a bit? 

![Image](https://github.com/user-attachments/assets/1c52cb53-72cf-4d89-b642-ce07f5e9a94b)

VaiTon:
>  Clicking on images to zoom on them is broken

@teolemon I cannot reproduce this

teolemon:
Let's assume it's fixed @VaiTon

PR #280, opened 2025-03-13
feat: add Robotoff questions

Abbas-Askari:
### What
Add support for robotoff questions.
Note: https://github.com/openfoodfacts/openfoodfacts-nodejs/pull/599 Need to be merged first. I have tested both PR locally.

### Screenshot
Not found:
image
`requestState = 'idle'`
image
`requestState = 'error'`
image
`requestState = 'loading'`
image
`requestState = 'answered'`
image
Light mode:
image
Responsiveness:
image

### Fixes bug(s)
- #264

PR #410, opened 2025-04-01
fix: Hide debug option for questions in product page(Prodution) 

KarruHarin:
### What
Rendered debug based on production or development.

### Part of 
- #337

### DIFF
--- a/src/routes/products/[barcode]/+page.svelte
+++ b/src/routes/products/[barcode]/+page.svelte
@@ -3,6 +3,7 @@
 	import { isConfigured as isPriceConfigured } from '$lib/api/prices';
 	import { isConfigured as isFolksonomyConfigured } from '$lib/api/folksonomy';
 	import { preferences } from '$lib/settings';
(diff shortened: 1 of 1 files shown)


github-throwaway:
Regardless of whether we want this change or not, I would expect an update in the readme which explains how to turn on and off development mode. And an update to the env example file.

KarruHarin:
> Regardless of whether we want this change or not, I would expect an update in the readme which explains how to turn on and off development mode. And an update to the env example file.

image
As you can see here the development mode for the debug options is not taken from .env rather from '$app/environment', so whenever you run it locally automatically the development mode is turned on .
In the same way debug options are hid in the other places :
For example:
In knowledge panels :
image

github-throwaway:
@VaiTon do you want this in a separate pull request.

VaiTon:
We already have #280 that is implementing this, I don't think merging this is worth the hassle of then rebasing that one.

Issue #421, opened 2025-04-02
Support for Connecting to a Locally Hosted Backend


akashshetty1997:
### Problem  
The backend channel does not support connecting to a locally hosted backend. This makes it harder to test and develop without deploying the backend remotely.  

### Proposed solution  
Allow the backend channel to connect to a locally hosted backend. Provide a way to configure this and document how it works.  

### Additional context  
Local development and testing would be easier with this feature. Right now, every change requires a remote deployment, which slows things down.

swastikCommits:
I would like to work on it.

VaiTon:
@akashshetty1997 we already have support for changing the API base URL via `.env` file. Is there something in particular that you aren't able to change with that file?

akashshetty1997:
I wanted to run a backend locally and connect the frontend to it, but I was informed this functionality isn't supported yet.

swastikCommits:
There is already `VITE_OFF_BASE_URL` in .env file, if you want a similar way to run the PricesApi locally #424 does this for you

swastikCommits:
@akashshetty1997 if it is so, then you can easily make changes to the .env to run it local host.
Sorry if you were misinformed.
I will be closing my corresponding PR for it, if you need any further assistance then let me know

akashshetty1997:
@swastikCommits thank you so much, yeah i believe there was misunderstanding.

swastikCommits:
@akashshetty1997 If you are satisfied with the feedback, I would urge you to close this issue

PR #424, opened 2025-04-03
feat: implementation of locally hosted Backend URL

swastikCommits:
### What
- Added support for a new environment variable `VITE_PRICES_API_LOCAL_URL`
- First we will try for `VITE_PRICES_API_LOCAL_URL`, if not available, then fallback to production URL i.e. `VITE_PRICES_API_URL`

### Fixes bug(s)
#421

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,6 @@
 VITE_FOLKSONOMY_API_URL=https://api.folksonomy.openfoodfacts.net
 VITE_OFF_BASE_URL=https://world.openfoodfacts.org
+VITE_PRICES_API_URL=https://prices.openfoodfacts.org
+
+# For local development with a locally hosted backend, uncomment and set this:
+# VITE_PRICES_API_LOCAL_URL=http://localhost:8000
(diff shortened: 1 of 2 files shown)


VaiTon:
Hey, sorry for the delay! Why can't I simply change the original `VITE_PRICES_API_URL` if I want to select another host?

VaiTon:
Also, I don't think that the issue was specifically talking about the Prices backend

swastikCommits:
> Hey, sorry for the delay! Why can't I simply change the original `VITE_PRICES_API_URL` if I want to select another host?

Yup as per the issue for local development it will first check for `VITE_PRICES_API_LOCAL_URL` if someone hasn't set it then it would fallback to `VITE_PRICES_API_URL`
Either we can do it. We can directly change `VITE_PRICES_API_URL` if you suggest, or for better checks we can fall back to the other URL

swastikCommits:
> Also, I don't think that the issue was specifically talking about the Prices backend

I'm so sorry I didn't meet your expectations. It would be my humble request can you please guide me approach the solution.

We can implement local development support for
- Main Open Food Facts API which we are currently using as `VITE_OFF_BASE_URL`
- Folksonomy API as `VITE_FOLKSONOMY_API_URL`

PR #433, opened 2025-04-04
bug: fixed some products that were not loading

### LINKED ISSUE #432: Some products not loading
### What
Some of the products are not loading like 
[https://openfoodfacts-explorer.vercel.app/products/5449000223555](https://openfoodfacts-explorer.vercel.app/products/5449000223555)
[https://openfoodfacts-explorer.vercel.app/products/3256220321372](https://openfoodfacts-explorer.vercel.app/products/3256220321372)

### Expected behavior
Solve the bug

Bhavishya2601:
Some products were not loading properly

Fixes #432

### DIFF
--- a/src/routes/products/[barcode]/GreenScore.svelte
+++ b/src/routes/products/[barcode]/GreenScore.svelte
@@ -39,7 +39,9 @@
 
 	let { grade }: { grade: string } = $props();
 	let normalizedGrade = $derived(grade.toLowerCase());
-	let { name, src, textColor, bgColor } = $derived(GRADE_MAP[normalizedGrade]);
+	let { name, src, textColor, bgColor } = $derived(
+		GRADE_MAP[normalizedGrade] ?? GRADE_MAP['unknown']
+	);
 </script>
 
 <div


src/routes/products/[barcode]/GreenScore.svelte
```diff
@@ -39,7 +39,9 @@
 
 	let { grade }: { grade: string } = $props();
 	let normalizedGrade = $derived(grade.toLowerCase());
-	let { name, src, textColor, bgColor } = $derived(GRADE_MAP[normalizedGrade]);
+	let { name, src, textColor, bgColor } = $derived(
+		GRADE_MAP[normalizedGrade] ?? GRADE_MAP['unknown']
```
Comment (VaiTon) [inline src/routes/products/[barcode]/GreenScore.svelte:43]: First, why do we have a normalizedGrade that is not handled by us?

What is the problem here? This is hiding the problem under a rug...

Comment (Bhavishya2601) [reply src/routes/products/[barcode]/GreenScore.svelte:43]: I found that for some products greenScore had a value of 'not-applicable' in them, which is not being handled by us currently
not-applicable should have same results as unknown, I thought maybe in this data, there maybe some products which return another value like not-applicable which if not present at that time, may again cause error

so I went with this way
If you want, instead of using this way, I can handle it in the GRADE_MAP

VaiTon:
@Bhavishya2601 this PR has a generic title and is a duplicate of #433. I'm closing this, sorry.

PR #442, opened 2025-04-07
feat: added organic preference in product page

### LINKED ISSUE #440: Implement the Organic product preference
Implement the organic product preference

Image

Bhavishya2601:
Added organic preference in product page

image

Fixes #440

### DIFF
--- a/src/lib/api/product.ts
+++ b/src/lib/api/product.ts
@@ -96,7 +96,7 @@ export class ProductsApi {
 		if (!res.ok) {
 			throw new Error(`Failed to fetch product attributes for barcode: ${barcode}`);
 		}
-	
+
 		return await res.json();
 	}
 
--- a/src/lib/api/product.ts
+++ b/src/lib/api/product.ts
@@ -89,6 +89,17 @@ export class ProductsApi {
 		return res.status === 200;
 	}
 
+	async getProductAttributes(barcode: string): Promise<any> {
+		const url = `${API_HOST}/api/v2/product/${barcode}?fields=product_name,code,attribute_groups_en`;
+		const res = await this.fetch(url);
+
+		if (!res.ok) {
+			throw new Error(`Failed to fetch product attributes for barcode: ${barcode}`);
+		}
+	
+		return await res.json();
+	}
+
 	async getProductReducedForCard(barcode: string): Promise<ProductState<ProductReduced>> {
 		const params = new URLSearchParams({
 			fields: REDUCED_FIELDS.join(','),
--- a/src/routes/products/[barcode]/+page.svelte
+++ b/src/routes/products/[barcode]/+page.svelte
@@ -194,7 +194,7 @@
 	</div>
 </Card>
 
-<div class="flex w-full justify-between flex-wrap gap-3 max-md:flex-col">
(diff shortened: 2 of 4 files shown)


swastikCommits:
Can't we wrap all the badges in a single row?

Bhavishya2601:
>Can't we wrap all the badges in a single row?

no, that would make it congested

this is what we'll get
image

swastikCommits:
Keeping just a single badge looks a bit off. Can you reduce the svgs size and wrap the text inside it. Wby?

Bhavishya2601:
>An alternative way would be keeping the badges in a 2×2 manner, similar to the main website.

the main website has 1 badge in new line
image


Keeping 2x2 would lead to unwanted gap

swastikCommits:
> Keeping 2x2 would lead to unwanted gap

Why will 2x2 create a gap? We will have a total of 4 badges after this implementation
.

Bhavishya2601:
>Why will 2x2 create a gap? We will have a total of 4 badges after this implementation

I meant to say if we keep the same width then that would lead to unwanted gap

but if we increase the gap, then it would be ok
image


@VaiTon @github-throwaway wdyt? which is better initial one or this one?

swastikCommits:
> > Why will 2x2 create a gap? We will have a total of 4 badges after this implementation
> 
> I meant to say if we keep the same width then that would lead to unwanted gap
> 
> but if we increase the width, then it would be ok image
> 
> @VaiTon @github-throwaway wdyt? which is better initial one or this one?

If we keep 2x2, then, ofc we need to increase the width.
Just a little bit of alignment in the text, and it would be good to go.
image

teolemon (changes requested):
We should be using https://wiki.openfoodfacts.org/Product_Attributes instead of hardcoding each setting

Bhavishya2601:
@teolemon I chose to hardcode each setting because rest of the preferences (nova, ecoscore, nutriscore) were also harcoded, no worries if you want me to use product_Attributes, I'll change it :)

swastikCommits:
> We should be using https://wiki.openfoodfacts.org/Product_Attributes instead of hardcoding each setting

This looks like a good issue to work on instead of harcoding Nutri-Score, Nova and Eco-Score. Shall I make a separate issue for it? @teolemon 
This one would be divided into 3 subtasks i.e. Nova, Nutri-Score and Eco-score. Wby @Bhavishya2601 ?

Bhavishya2601:
@swastikCommits yeah we can work on that after merging this PR?

@teolemon @VaiTon

VaiTon:
Let's work towards https://wiki.openfoodfacts.org/Product_Attributes

Bhavishya2601:
>Let's work towards https://wiki.openfoodfacts.org/Product_Attributes

@VaiTon it was implemented according to this only, can I pls know the reason why you closed this pr?

VaiTon:
Because you still created a specialized component to handle the organic field. Handling product attributes should be done in a general manner, like Knowledge panels

Bhavishya2601:
ok, do you want to discuss the implementation in the issue #446 and then proceed

PR #448, opened 2025-04-08
fix: text-wrapping in folksonomy engine

### LINKED ISSUE #447: Text-wrapping in the Folksonomy Engine visualisation
### Story
- Text-wrapping in the Folksonomy Engine visualisation
Image
- the current implementation is not practical on desktop: https://openfoodfacts-explorer.vercel.app/products/3017620422003

Bhavishya2601:
### What
text wrapping issue in folksonomy engine

image

As the input field does not support text wrapping, so we are gonna toggle between input and textarea fields based on if the user is logged in or not, as the user gets write access if the user is logged in

Fixes #447

### DIFF
--- a/src/routes/products/[barcode]/Folksonomy.svelte
+++ b/src/routes/products/[barcode]/Folksonomy.svelte
@@ -112,13 +112,21 @@
 					</div>
 				</td>
 				<td class="flex gap-2" aria-label="Value">
-					<input
-						type="text"
-						class="input grow max-sm:w-20"
-						value={tag.v}
-						readonly={!loggedIn}
-						onchange={(e) => updateTag(e.currentTarget.value, i)}
-					/>
+					{#if loggedIn}
+						<input
+							type="text"
+							class="input grow max-sm:w-20"
+							value={tag.v}
+							readonly={!loggedIn}
+							onchange={(e) => updateTag(e.currentTarget.value, i)}
+						/>
+					{:else}
+						<textarea
+							class="textarea grow max-sm:w-20"
+							readonly
+							style="white-space: pre-wrap; overflow-wrap: break-word;">{tag.v}</textarea
+						>
+					{/if}
 				</td>
 				{#if loggedIn}
 					<td>
--- a/src/routes/products/[barcode]/Folksonomy.svelte
+++ b/src/routes/products/[barcode]/Folksonomy.svelte
@@ -112,18 +112,12 @@
 					</div>
 				</td>
 				<td class="flex gap-2" aria-label="Value">
-					{#if loggedIn}
-						<input
-							type="text"
-							class="input grow max-sm:w-20"
-							value={tag.v}
-							onchange={(e) => updateTag(e.currentTarget.value, i)}
-						/>
-					{:else}
-						<textarea class="textarea grow break-words whitespace-pre-wrap max-sm:w-20" readonly
-							>{tag.v}</textarea
-						>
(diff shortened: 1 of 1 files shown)


src/routes/products/[barcode]/Folksonomy.svelte
```diff
@@ -112,13 +112,21 @@
...
 				</td>
 				<td class="flex gap-2" aria-label="Value">
-					<input
-						type="text"
-						class="input grow max-sm:w-20"
-						value={tag.v}
-						readonly={!loggedIn}
-						onchange={(e) => updateTag(e.currentTarget.value, i)}
-					/>
+					{#if loggedIn}
```
Comment (VaiTon) [inline src/routes/products/[barcode]/Folksonomy.svelte:115]: I don't think this if is related with the title nor the description of this PR?

VaiTon (changes requested):

src/routes/products/[barcode]/Folksonomy.svelte
Comment (Bhavishya2601) [reply src/routes/products/[barcode]/Folksonomy.svelte:115]: As the input field does not support text wrapping, so in order to wrap the text we need to use textarea

The if condition is to check if the user is logged in or not, based on it input field or textarea will be shown

Bhavishya2601:
@VaiTon This PR is ready to be reviewed, You can let me know if you need any changes

src/routes/products/[barcode]/Folksonomy.svelte
Comment (VaiTon) [reply src/routes/products/[barcode]/Folksonomy.svelte:115]: Still I don't like that this is done in a PR that has no mention of this. As a good rule of thumb, if I see something in the diff that is not mentioned in the description I'll block the PR

src/routes/products/[barcode]/Folksonomy.svelte
```diff
@@ -112,13 +112,21 @@
...
+						<input
+							type="text"
+							class="input grow max-sm:w-20"
+							value={tag.v}
+							readonly={!loggedIn}
```
Comment (VaiTon) [inline src/routes/products/[barcode]/Folksonomy.svelte:120]: This can never be true

src/routes/products/[barcode]/Folksonomy.svelte
```diff
@@ -112,13 +112,21 @@
...
+							onchange={(e) => updateTag(e.currentTarget.value, i)}
+						/>
+					{:else}
+						<textarea
+							class="textarea grow max-sm:w-20"
+							readonly
+							style="white-space: pre-wrap; overflow-wrap: break-word;">{tag.v}</textarea
```
Comment (VaiTon) [inline src/routes/products/[barcode]/Folksonomy.svelte:127]: Do not use inline style if possible. Use Tailwindcss

VaiTon (changes requested):

src/routes/products/[barcode]/Folksonomy.svelte
```diff
@@ -112,13 +112,18 @@
...
-						readonly={!loggedIn}
-						onchange={(e) => updateTag(e.currentTarget.value, i)}
-					/>
+					{#if loggedIn}
+						<input
+							type="text"
+							class="input grow max-sm:w-20"
+							value={tag.v}
+							onchange={(e) => updateTag(e.currentTarget.value, i)}
+						/>
```
Comment (VaiTon) [inline src/routes/products/[barcode]/Folksonomy.svelte:121]: Why the logged in users don't get the benefit of a textarea spanning multiple lines?

Comment (Bhavishya2601) [reply src/routes/products/[barcode]/Folksonomy.svelte:121]: You are right, Let me change it :)

VaiTon (approved):

Issue #426, opened 2025-04-04
Inconsistency of products design in landing page


swastikCommits:
### What
After the merge of a new PR, the products look off in design.

### Current behavior
1. The padding in the left and right of the products, makes it look unsymmetrical with the current homepage.
2. The loader/skeleton is very small as compared to the the actual products UI which looks inconsistent.
3. The loader when we click on a specific product is shifted upwards, while earlier it used to be in the center.
4. The product images are stretched in some of the cases.

### Screenshots

1. The padding need to be removed.

![Image](https://github.com/user-attachments/assets/e539e50e-3698-4f6d-bcb1-06f85728b766)

Mobile version: 

![Image](https://github.com/user-attachments/assets/7ef87882-3947-4fb1-85a8-f2644f8c27ed)

2. Loader/skeleton should look similar in dimensions as the products container

![Image](https://github.com/user-attachments/assets/aeabfe95-d014-4acd-82b4-5576575fd9de)

3.  
![Image](https://github.com/user-attachments/assets/4f865d18-3dea-4b49-8469-51faa3ba67f4)

4. 
![Image](https://github.com/user-attachments/assets/9b7b59f8-f229-4e83-924b-3d379b87320c)

### Platform (Desktop, Mobile, Hunger Games)
 - OS: Desktop 
 - Platform: Brave Browser(desktop)

swastikCommits:
@VaiTon @github-throwaway What are your views on this? 
I will proceed after you approve.

swastikCommits:
Most probably this is due to #347

github-throwaway:
Please provide a mockup of your solution first. :)

github-throwaway:
I would think that #399 is somewhat related to this.

swastikCommits:
Sure, I will be providing a mockup of the solution soon.
As of #399 it should follow up to sync with the landing page design

swastikCommits:
In my opinion #347 was concerned about design of landing page and #399 for the search results page, as the PR which got merged left some UX issues which should be fixed here(after maintainers approve the mockup solution)
Followed by fixing the landing page, the search results page must be updated(#399 )

github-throwaway:
I would expect these two places (search results and landing page) to use the same logic and layout.

swastikCommits:
Yup exactly! 💯

swastikCommits:
[...]

PR #454, opened 2025-04-09
style: fixed the inconsistency of UI in landing page

swastikCommits:
### What
I tried to use the minimal amount of absolute values in classes, while keeping it responsive to all screens.
Everyone are invited to share their valuable suggestions and insights. 

### Screenshot

1.Screenshot 2025-04-09 171520
Screenshot 2025-04-09 171754

2. 
Screenshot 2025-04-09 171533

3. 
Screenshot 2025-04-09 171557

### Fixes bug(s)
#426

### DIFF
--- a/src/lib/ui/SmallProductCard.svelte
+++ b/src/lib/ui/SmallProductCard.svelte
@@ -24,13 +24,15 @@
 
 <a
 	href={`/products/${product.code}`}
-	class="flex justify-center sm:m-2 sm:w-64"
+	class="flex w-full justify-center"
 	class:pointer-events-none={navigating.to}
 >
-	<div class="dark:bg-base-200 text-primary flex h-40 w-80 rounded-2xl shadow-md">
-		<div class="h-full w-32 text-center text-sm sm:w-28">
+	<div class="dark:bg-base-200 text-primary flex h-40 w-full max-w-full lg:max-w-xl rounded-2xl shadow-md">
+		<div class="h-full w-32 text-center text-sm sm:w-28 lg:w-36">
 			{#if navigating.to?.params?.barcode === product.code}
-				<span class="loading loading-ring loading-lg mx-auto my-auto"></span>
+				<div class="flex h-full w-full items-center justify-center">
+					<span class="loading loading-ring loading-lg"></span>
+				</div>
 			{:else if product.image_front_small_url}
 				<img
 					src={product.image_front_small_url}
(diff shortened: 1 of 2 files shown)


Bhavishya2601:
> Screenshot 2025-04-09 171533

@swastikCommits How about we display 9 of these (same count as when the product loads), so that when page loads it doesn't change the page height

And can you also make them the same height as of the product card

swastikCommits:
> I found currently it has a value of rounded-2xl, I tested it with rounded-lg, I think so that's much better, wdyt?

@Bhavishya2601 may I know how did you find `rounded-lg` better than `rounded-2xl`?

Here's a closer look:
`rounded-lg`: 
Screenshot 2025-04-09 205639


`rounded-2xl`: 
Screenshot 2025-04-09 205722

swastikCommits:
Yes, we can increase the number of skeleton cards @Bhavishya2601 , but there's a catch: we can't always fetch 9 product cards. 
I have kept it to the default '4' because that seems to be the minimum number of product cards we can fetch from the BE.

But I would want to know @github-throwaway and @VaiTon pov on it too!

abrahamvg:
Hi @swastikCommits 
Since you are working on this, would like to provide my suggestions:
- Keep the main container width to 80% of the viewing width because currently it is too much for desktop view and too less for mobile view (or alt keep padding left right to 10%)
- The skeleton count currently is 4 during page load, propose to make it 9 (I quite didn't understand with the statement `we can't always fetch 9 product cards` when the api payload sent itself is for count 10 so would like to understand this further) and similiar to the size of the card so that it is symmetric and represents the actual card.
- Increase the gap between each cards because currently it is quite less, again something like 12px to 14px would work fine.
-  It would be great if you can make a small visual hierarchy for the below 
image
I actually didn't understand what LU means here until I saw in the `api` and that it is the name of the brand. I suggest to keep the text hierarchy something like this :
image

Bhavishya2601:
@swastikCommits you changed the border radius for the card, change it for the the image as well and then the corners won't peak out 
the images also has a border radius 2xl on them

> 
> @Bhavishya2601 may I know how did you find `rounded-lg` better than `rounded-2xl`?
> 
> Here's a closer look: `rounded-lg`: Screenshot 2025-04-09 205639
> 
> `rounded-2xl`: Screenshot 2025-04-09 205722

heymitali:
Hey @swastikCommits, I have a couple of  improvement suggestions for PR.
1-  In the second screenshot for mobile screens, there is a lot of empty space on the right side. I believe that the card's content must be uniformly displayed in the card container. 

Screenshot 2025-04-10 at 1 08 55 AM


 2- In the desktop screens, we must maintain some margins around the cards. Right now, its looking a lot cluttered in the screenshot you provided.

Screenshot 2025-04-10 at 1 08 36 AM

VaiTon (approved):
Thank you!

swastikCommits:
Thanks for the suggestion @heymitali 
I know those badges are placed closely; I tested it keeping them apart, but they didn't seem visually good at all.
Can't help now that it got merged :)

Issue #398, opened 2025-03-30
Alignment issue while rotating product image


Bhavishya2601:
### What
When we try to rotate the product image, it overlaps other thinks like buttons, heading etc as shown in video below

https://github.com/user-attachments/assets/0c0ec73a-57ed-470c-a28e-382df0403940

### Expected behavior
I think so a much better way would be to add a maximize button instead of the rotate button which will magnify the image bringing it forward and then we can rotate it when it is in full screen

swastikCommits:
So how are you gonna approach with this?
If you think decreasing the dimensions of the image to avoid overlap then ig it won't be a great idea

swastikCommits:
If someone wants to rotate the image he/she will have to negotiate on something or else decreasing the image dimensions will create visibility issues for the user who doesn't want to rotate the image.

swastikCommits:
So I would request to @VaiTon to approve this issue with approach solution to fix it

Bhavishya2601:
>So I would request @VaiTon to approve this issue with the corresponding approach to fix it

Bro, I found this a little disrespectful 
I appreciate your concern over the issue, but I obv planned it before creating the issue

Anyways let's get to the point

So the current implementation of the image isn't friendly, it is overlapping other text, so what I was thinking is we can add a icon for going to full screen mode or just by clicking the image
after the user clicks the image, the image would go into full screen mode, something like this

![Image](https://github.com/user-attachments/assets/9c37c752-3be5-49df-979f-8fe7749fde9b)

And then when in full screen we will provide user with the option of rotating the image and a button which would exit the full screen mode (if you want anything else to be added then we can add that as well)

this approach would solve the current image of overlapping the text and would also be user friendly

swastikCommits:
I apologise for my words, I didn't mean to hurt you. 
I got your point but above working on the Image modal. This would be meaningful to work after the Image modal is fixed in the respective PR(a zoom in and zoom out function is also added on top of it)

Bhavishya2601:
No worries about the PR bro, I'll wait till it's merged

I created this issue for the mentors to review the idea

swastikCommits:
Thanks for the update @Bhavishya2601, again I'm really sorry if any of my words hurted you. I never meant it.

VaiTon:
> I think so a much better way would be to add a maximize button instead of the rotate button which will magnify the image bringing it forward and then we can rotate it when it is in full screen

[...]

PR #476, opened 2025-04-14
fix: product image overflows from the container

swastikCommits:
## Description

- Previously, product images would overflow from their container.
- They are now fixed inside the container while taking care that they don't get clipped by the contents surrounding that container while rotating the product image.
- The design has been made responsive for both mobile and large/medium screens.

## Screenshots
image

This image now doesn't get clipped on rotation
Screenshot_2025_0414_170400

---

## Checklist:

**Author Self-Review:**


Bhavishya2601:
@swastikCommits We can't be having that small image of the product on the page, I'm barely able to see the product and the rest of the space around the image is going to waste

image

Well the issue that you linked (#398), I think we were discussing a different approach to rotate the image by bringing it forward, not to make the product image small so that it doesn't overlap

wdyt @VaiTon @github-throwaway ??

swastikCommits:
Yeah ikr, just in case for review, I had added `overflow-hidden`. If the maintainers won't accept such small image, then I will remove that class. This issue of overlapping of the image of other elements would be further fixed in the upcoming PR in a better way- "Zoom Functionality"(I'm still working on it)

This PR's main concern was to fix the product image by default used to overflow from the container.

Bhavishya2601:
>Fixes https://github.com/openfoodfacts/openfoodfacts-explorer/issues/398

Can you pls unlink the issue #398 from your initial msg, I don't think this PR solves that issue, we are discussing a different approach there and the discussion is still going on

>This PR's main concern was to fix the product image by default used to overflow from the container.

From my perspective, I don't think this PR makes sense, it like hiding the issue under a rug
And moreover @github-throwaway did suggest in the issue #398 to move the rotate image option to edit screen

swastikCommits:
In short: 
1. Previously by default the product image used to overflow from downwards of the container.
2.  At the same time some images on rotation overflow from the container from topwards overlapping on "Edit" button.
Here this PR's main job was to fix 1 not 2(this is fixed too in this PR)
In case @VaiTon and @github-throwaway don't get satisfied by the solution of 2, I'll make changes and this PR will primarily focus on fixing 1.

Bhavishya2601:
I don't wanna get in an argument or something, it's mentors call now

I just gave my viewpoint

swastikCommits:
> > Fixes #398
> 
> Can you pls unlink the issue #398 from your initial msg, I don't think this PR solves that issue, we are discussing a different approach there and the discussion is still going on
> 
> > This PR's main concern was to fix the product image by default used to overflow from the container.
> 
> From my perspective, I don't think this PR makes sense, it like hiding the issue under a rug
> 
> And moreover @github-throwaway did suggest in the issue [#398 (comment)](https://github.com/openfoodfacts/openfoodfacts-explorer/issues/398#issuecomment-2793926723) to remove the rotate image option from product page and move it to edit page
Nothing's been hidden under the rug, FYI both the issues were regarding overflowing from the container. I have dug deep into all classes fixing it, and tested it extensively for several product images.
So, be patient let the maintainers review, if they don't like it, I will make further changes to fix the primary issue only.

src/lib/ui/ImageButton.svelte
```diff
@@ -15,16 +15,18 @@
...
 		{#if src != null}
-			<div class="flex items-center justify-center">
-				<img
-					{src}
-					{alt}
-					class="max-h-full max-w-full rounded-lg object-contain"
-					style="transform: rotate({rotation}deg); transition: transform 0.3s ease;"
-				/>
+			<div class="flex h-full w-full items-center justify-center p-4 md:py-8">
+				<div class="relative" style="width: 75%; height: 75%;">
```
VaiTon, line 22: No inline style. Also, why do we need this width and height capped to 75%?

src/lib/ui/ImageButton.svelte
```diff
@@ -15,16 +15,18 @@
 </script>
 
 <ImageModal bind:this={modal} />
-<div class="relative">
-	<button class="flex max-w-full justify-center" {onclick}>
+<div class="relative flex h-full w-full items-center justify-center">
+	<button class="flex h-full w-full items-center justify-center" {onclick}>
```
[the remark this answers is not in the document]
VaiTon: Is it?

VaiTon (changes requested):

src/lib/ui/ImageButton.svelte
swastikCommits: @VaiTon by deault if we rotate the images some of them will not overlap on other elements while rotating, while some will still overlap(for example, if it is a more of a vertical image). So I figured out that, capping its dimensions did help.
This one was just an experimentation for your review, as this reduces the size of product image.

github-throwaway:
A vertical image will be at its tallest. So the vertical space available to the image to not overlap with anything. must be at least its largest dimension (either width or height). Why don’t we make sure that there is at least this much space?

with this being said: I would move the entire image rotation to the edit screen. To me it’s not clear what the rotation does. Is it a local setting? Do I write that information back to the database?

Moving that to the edit screen would make it clearer.

swastikCommits:
@github-throwaway Rotation of images is done in products page so that users can rotate it be and be clear about the image's information, as the product images recieved from the backend are not always in the perfect manner to view(eg. some images uploaded in BE are vertical un nature). 
Putting it to the edit screen might not add value to it as it just for better user experience, no changes are made to the original image in the database after rotation.
A better option would be moving it to the ImageModal which will be enhanced in the PR #483

swastikCommits:
To be precise, I will further make changes to specifically address the issue of the image overflowing from the container by default (without even rotation):
Screenshot 2025-04-16 130733

github-throwaway:
If that’s your intended use case then I would only show the rotate buttons when a user click on the picture. And it gets shown almost full screen. Then show left and right rotating options. Otherwise the image is too small. In my opinion.

swastikCommits:
Yeah exactly we'll be moving the rotate button to ImageModal i.e. the rotation feature will only be available after the image is clicked.
This use case will be addressed under the issue #398

swastikCommits:
I messed up while solving the merge conflicts. Maybe'll raise this a new PR with the new changes:(

PR #483, opened 2025-04-16
feat: implementation of Zoom functionality on product Images

### LINKED ISSUE #329: feat: Add Zoom Function to Product Images on Products Page
### Problem
The products page currently displays product images but lacks a zoom function, limiting users' ability to closely inspect details (e.g., nutrition labels, ingredients). Adding a zoom feature would enhance usability and accessibility, especially for users needing to examine fine print or verify product information.

### Proposed solution
- Users should be able to zoom in/out on product images on the products page.
- Zoom can be triggered by hover, click, or a dedicated button (e.g., magnifying glass icon).
- Smooth scaling with a maximum zoom level (e.g., 2x or 3x) and ability to pan if the image exceeds the viewport.
- Responsive design to work on both desktop and mobile.

### Part of
None

swastikCommits:
## Description
- Implemented user-friendly zoom functionality for product images.  
- Shows the current zoom level with consistent styling, including Zoom In, Zoom Out, and Reset Zoom buttons.  
- Includes a full-screen modal option that displays the image at the centre.

https://github.com/user-attachments/assets/45640690-d948-47f5-96e3-18bc1c84653b

Fixes #329 

---

## Checklist:

**Author Self-Review:**


swastikCommits:
@VaiTon can you please review this PR?

swastikCommits:
@VaiTon can you please review this PR?

Bhavishya2601:
Btw do we really need this? I don't think any user would zoom that in on an image, isn't just bringing image to full screen enough?

VaiTon:
@Bhavishya2601 sometimes it's needed if images are low quality

VaiTon (commented):
I see that you added the `className` prop but I can't see where it should be useful. If it was added for future proofing the component I would remove it. I don't really like the React approach and I think every component should have (when possible) its logic for its style.

swastikCommits:
Yup I used className for future proofing. If you didn't like this approach, I'll remove it, as it won't affect the functionality.

src/lib/ui/ZoomableImage.svelte
```diff
@@ -0,0 +1,89 @@
...
+	let { src, alt }: Props = $props();
+
+	let zoomLevel = $state(1);
+	const MAX_ZOOM = 3;
+
+	function zoomIn(e: MouseEvent) {
+		e.preventDefault();
+		e.stopPropagation();
+		const newZoom = Math.min(zoomLevel + 0.5, MAX_ZOOM);
+		setZoomImageState({ currentZoom: newZoom });
```
VaiTon, line 15: You should `setZoomImageState` inside an `$effect` block and only edit `zoomLevel`, so that we always have a single source of truth

VaiTon (changes requested):

src/lib/ui/ZoomableImage.svelte
swastikCommits: I had removed the `setZoomImageState` calls from the zoom functions, and these functions will now update `zoomLevel`
Hence, I have added an `$effect` block that will watch for changes on `zoomLevel` and update the state accordingly.

src/lib/ui/ImageModal.svelte
```diff
@@ -11,27 +11,55 @@
 		| undefined = $state();
 
 	let dialog: HTMLDialogElement | undefined = $state();
-	let rotation = $state(0);
+	let zoomLevel = $state(1);
```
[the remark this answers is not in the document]
VaiTon: You should do the same $effect thing here.

VaiTon (commented):
Also, what is the relation between ImageModal and ZoomableImage? Why they both have a zoomLevel?

swastikCommits:
To be specific ZoomableImage isn't used now, as in the earlier follow up PR where I had added an additional feature where we used this component to zoom the product image without opening its ImageModal, which was ultimately disliked by you. After all I removed the imports but forgot to delete the file-`ZoomableImage.svelte`. 
Thanks for pointing out, and I apologize for the mistake :)

src/lib/ui/ImageModal.svelte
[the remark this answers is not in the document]
swastikCommits: Have a look made the changes

src/lib/ui/ImageModal.svelte
```diff
@@ -11,60 +11,123 @@
...
+		e.preventDefault();
+		e.stopPropagation();
+		zoomLevel = 1;
+	}
+
+	let container: HTMLDivElement | undefined = $state();
+	const { createZoomImage, setZoomImageState } = useZoomImageWheel();
+
+	$effect(() => {
+		if (setZoomImageState) {
```
VaiTon, line 50:
```suggestion
		if (setZoomImageState != null) {
```

src/lib/ui/ImageButton.svelte
```diff
@@ -5,7 +5,13 @@
...
 
 	let modal: ImageModal;
-	let onclick = $derived(src != null ? () => modal.displayImage(src, alt) : undefined);
+
+	function getFullSizeImageUrl(url: string): string {
+		return url.replace(/\.400\.|\.200\.|\.100\./, '.full.');
+	}
+	let onclick = $derived(
+		src != null ? () => modal.displayImage(getFullSizeImageUrl(src), alt) : undefined
+	);
```
VaiTon, line 14:
```suggestion
	let onclick = $derived.by(() => {
	    if (src == null) { return undefined }
	    
		return () => modal.displayImage(getFullSizeImageUrl(src), alt);
	});
```

VaiTon (approved):

PR #497, opened 2025-04-19
fix: fixed header issues on smaller screens

### LINKED ISSUE #496: Header responsiveness issues on Smaller screens
### What

-  The header is not responsive on smaller screens
- Extra padding on smaller screens in 'Welcome to OFF section'

Image

Bhavishya2601:
## Description

- The header was not responsive on smaller screens
- 'Welcome to OFF' has double padding

Before
image

After

https://github.com/user-attachments/assets/d4ac6369-1362-4573-92e2-57eaf94a379a

Fixes #496 

---

## Checklist:

**Author Self-Review:**

**LLM Usage Disclosure:** not used

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -108,13 +108,13 @@
 </div>
 
 <div class="bg-base-100 top-0 right-0 left-0 z-50 mx-4 xl:hidden">
-	<div class="navbar bg-base-100 mx-auto mt-2 mb-2 max-w-7xl">
+	<div class="navbar bg-base-100 mx-auto mt-2 mb-2 px-0">
 		<div class="navbar-start">
 			<a href="/">
 				<Logo />
 			</a>
 		</div>
-		<div class="navbar-end flex gap-2">
+		<div class="navbar-end flex gap-1 sm:gap-2">
 			<button
 				aria-label={$_('search.button')}
 				class="btn btn-square btn-secondary text-lg"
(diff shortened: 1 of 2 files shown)


src/routes/+layout.svelte
```diff
@@ -108,13 +108,13 @@
 </div>
 
 <div class="bg-base-100 top-0 right-0 left-0 z-50 mx-4 xl:hidden">
-	<div class="navbar bg-base-100 mx-auto mt-2 mb-2 max-w-7xl">
```
VaiTon, line 111: This class right here makes the navbar have a max width when the screen is big enough. We need it

src/routes/+layout.svelte
```diff
@@ -108,13 +108,13 @@
...
+	<div class="navbar bg-base-100 mx-auto mt-2 mb-2 px-0">
```
VaiTon, line 111: Why?

src/routes/+layout.svelte
```diff
@@ -108,13 +108,13 @@
...
 		<div class="navbar-start">
 			<a href="/">
 				<Logo />
 			</a>
 		</div>
-		<div class="navbar-end flex gap-2">
+		<div class="navbar-end flex gap-1 sm:gap-2">
```
VaiTon, line 117: Why?

VaiTon (changes requested):

src/routes/+layout.svelte
Bhavishya2601: navbar class is giving a padding of 0.5rem (8px), so px-0 is used for overriding it

src/routes/+layout.svelte
Bhavishya2601: To get a little more space between logo and btn's

with `gap-2`
image

with `gap-1 sm:gap-2`
image

Just a minor difference but I think so it looks good with gap-1, you can give me, your pov on this

src/routes/+layout.svelte
Bhavishya2601: @VaiTon I just checked, isn't this navbar for smaller devices which gets hidden after `xl` by property `xl:hidden`, I think the  `max-w-7xl` is placed in wrong place, it should be in the navbar which is for larger devices

correct me if I'm wrong

VaiTon: That's right, thanks!

VaiTon (approved):

PR #504, opened 2025-04-19
Improve error handling for network connectivity issues

akashshetty1997:
## Description

This PR improves how the app handles network connectivity issues, particularly when loading product data. The changes:

1. Now reliably identifies when the app can't reach Open Food Facts servers (including offline states and DNS failures)
2. Shows helpful instructions when network problems occur
3. Preserves all existing error handling while adding these improvements
4.  Added new error messages in:
     - `en.json`
     - `it.json` 

Key changes:
- Enhanced `+page.ts` to properly catch and classify network-related failures
- Updated the error page to display user-friendly network error messages
- Added status page reference for troubleshooting

https://github.com/user-attachments/assets/26c79d93-da37-4c59-a084-026a13c7879c

Fixes #[287](https://github.com/openfoodfacts/openfoodfacts-explorer/issues/287)

## Checklist:

**Author Self-Review:**

- [x] I have performed a self-review of my own code
- [x] I understand the changes I'm proposing and why they are needed
- [x] My changes generate no new warnings or errors (linting, console)
- [ ] I have made corresponding changes to the documentation (if applicable)

**LLM Usage Disclosure:**

- [x] I used an AI Large Language Model to assist with implementation, and have manually verified all generated code for accuracy and security

**Triggering Code Review:**

- You can request an AI-powered code review by commenting `/gemini review` on this PR after it's been created

### DIFF
--- a/src/routes/+error.svelte
+++ b/src/routes/+error.svelte
@@ -4,6 +4,7 @@
 
 	let message = $derived(page.error?.message);
 	let errors = $derived(page.error?.errors);
+	let isNetworkError = $derived(message === 'network_error');
 
 	$effect(() => {
 		for (const error of errors ?? []) {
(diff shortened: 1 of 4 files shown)


src/routes/+error.svelte
```diff
@@ -16,42 +17,57 @@
...
-									<span>
-										<code>{error.field.id}</code> has value
-										<code>{error.field.value.length != 0 ? error.field.value : '<empty>'}</code>
-									</span>
-								{/if}
-							</ul>
-						</li>
-					{/each}
-				</ul>
+		{#if isNetworkError}
```
VaiTon, line 24: Can we extract the two if branches to Svelte snippets?

src/routes/+error.svelte
```diff
@@ -16,42 +17,57 @@
...
+			<h1 class="mb-3 text-xl font-bold">{$_('errors.network.title')}</h1>
+			<p class="mb-4 text-sm">
+				Failed to connect to the Open Food Facts server. Please check your internet connection and
+				try again.
```
VaiTon, line 28: This should be i18n-able

src/routes/+page.ts
```diff
@@ -19,46 +20,75 @@ const count = '10';
...
+		const productsPromises = response.questions.map((question) =>
+			productApi.getProductReducedForCard(question.barcode)
+		);
 
-	return Promise.all(productsPromises);
+		return Promise.all(productsPromises);
+	} catch (err: any) {
+		console.error('productsWithQuestions error:', err);
+		throw err;
+	}
```
VaiTon, line 38: Why should we catch an exception just to rethrow it?

VaiTon (changes requested):
Thank you! Just a few nitpicks

src/routes/+error.svelte
akashshetty1997: will do that

src/routes/+error.svelte
akashshetty1997: I'll create two separate components:

- NetworkError.svelte (for connection issues)
- StandardError.svelte (for else condtion)

Originally I kept them together because:

- This error UI is only used in one place
- The logic was closely tied to this page

But I agree separating them is better because:

- It makes the code more organized
- Each component focuses on one responsibility
- Easier to maintain and modify independently

src/routes/+page.ts
akashshetty1997: Agreed, the catch block doesn't make sense here.

akashshetty1997:
I split the code into components for better organization. For the i18n changes, initialization timing was improved by using waitLocale to properly load translations before rendering components. This fixed the "cannot format error" that occurred when components tried to access translations before they were ready.

src/lib/i18n/index.ts
```diff
@@ -12,13 +12,24 @@ locales.forEach((locale) => {
 	});
 });
 
-export function initI18n() {
+init({
+	fallbackLocale: 'en',
+	initialLocale: 'en'
+});
```
VaiTon, line 18: Why do we do this outside the function? I think this makes every file that imports i18n run this function

VaiTon:
@akashshetty1997 can we split the i18n changes to a different PR?

src/lib/i18n/index.ts
akashshetty1997: I added that initial setup outside the function because our app was crashing with that 'Cannot format message' error whenever i  switched languages to Italian. Basically, components were trying to use translations before they were ready.
So this fix makes sure there's always at least English available right from the start, no matter what. The regular initI18n() function still handles setting up the user's preferred language, but this prevents those  crashes during language switching.
I get your point about it running on every import, but in this case that's actually helpful  it guarantees translations are always available when needed, no matter when components try to use them.

akashshetty1997:
I'll revert the i18n initialization changes in this PR and create a separate PR for the i18n fixes.

Bhavishya2601:
@akashshetty1997 your checks are failing

akashshetty1997:
@Bhavishya2601  thnaks i fixed it.

src/lib/ui/StandardError.svelte
```diff
@@ -0,0 +1,51 @@
...
+		impact: { id: string };
+		message: { id: string };
+		field?: { id: string; value: string };
+	}> = [];
+
+	export let message: string = '';
+
+	function isError(error: string) {
+		return errors?.some((e) => e.message.id === error) ?? false;
+	}
```
VaiTon, line 14: Old Svelte v4 syntax. Can we convert to v5?

src/routes/+page.ts
```diff
@@ -20,45 +21,75 @@ async function productsWithQuestions(
...
+			err instanceof TypeError &&
+			(err.message === 'fetch failed' ||
+				err.message.includes('ENOTFOUND') ||
+				err.message.includes('Failed to fetch') ||
+				err.message.includes('NetworkError') ||
+				err.name === 'AbortError' ||
+				err.message.includes('timeout'));
+		if (isNetworkError) {
+			throw error(503, {
+				message: 'network_error',
```
VaiTon, line 65: Can we group these messages in some consts, and then use them in .svelte code?

VaiTon (changes requested):

src/routes/+page.ts
akashshetty1997: Okay I will Centralized all error handling in a new src/lib/errors.ts file that contains error messages, types, and status codes. This single source of truth eliminates duplicated strings across the codebase and makes future updates simpler. Also moved the network error detection logic into a reusable function that makes our code cleaner and more maintainable.  is this correct approach @VaiTon ?

VaiTon: Yes, and thanks!

VaiTon (approved):
LGTM, we can iterate on main

PR #505, opened 2025-04-20
feat: added VITE_API_VERSION parameter to test the API in dev mode

### LINKED ISSUE #401: Add a numeric field in dev mode to be able to use an arbitrary version of the API
### Context
- As of last week, there was a revamp of the API with more fine grained versioning
- For instance, changes in behaviour for Brands are behind a new version
- Add a numeric field in dev mode to be able to use an arbitrary version of the API, so that we can easily test behaviors before bumping the default version we use in the openfoodfacts-explorer, and so that the dev team can test behaviors on their local instances.
- https://github.com/openfoodfacts/openfoodfacts-server/blob/main/docs/api/ref-api-and-product-schema-change-log.md

swastikCommits:
## Description

_Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context._

Added `VITE_API_VERSION` env variable to `.env.example` with a default value of 3.
Updated the `PRODUCT_URL` function to use the API version from env variables.
Updated the `getProductAttributes` method in `ProductsApi` class to use the API version parameter.

**@VaiTon We must update the README to address the new API version configuration option which would be helpful for the developers to test it.**

Screenshot 2025-04-20 114443

Fixes #401 

---

## Checklist:

**Author Self-Review:**

**LLM Usage Disclosure:**

- [ ] If I **did** use an AI Large Language Model, I have reviewed the generated code/text to ensure its accuracy, security, and relevance to the project's context and licensing.

**Triggering Code Review:**

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,3 @@
 VITE_FOLKSONOMY_API_URL=https://api.folksonomy.openfoodfacts.net
 VITE_OFF_BASE_URL=https://world.openfoodfacts.org
+VITE_API_VERSION=3
--- a/src/lib/api/product.ts
+++ b/src/lib/api/product.ts
@@ -11,7 +11,8 @@ export class ProductsApi {
 	}
 
 	async getProductAttributes(barcode: string): Promise<ProductAttribute[]> {
-		const url = `${API_HOST}/api/v2/product/${barcode}?fields=product_name,code,attribute_groups_en`;
+		const apiVersion = import.meta.env.VITE_API_VERSION || '2';
+		const url = `${API_HOST}/api/v${apiVersion}/product/${barcode}?fields=product_name,code,attribute_groups_en`;
 
 		const res = await this.fetch(url);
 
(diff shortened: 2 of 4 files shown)


swastikCommits:
I have updated the `README.md` by adding a section called "Development Configuration". 
Let me know if that needs any further changes.

VaiTon:
@swastikCommits this is not what the issue was about. The api version is in general different from the product schema (we're at api v3 and product schema v 1001, see [think link](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/docs/api/ref-api-and-product-schema-change-log.md) for details)

swastikCommits:
Ah, thanks for correcting. 
That means when a client makes a request with a specific API version, the server converts the response to match the corresponding product schema version. So, we'll need to maintain the VITE_API_VERSION and handle the conversion between API versions and schema versions by adding another parameter that supports schema version.

swastikCommits:
As I tested it locally and added parameters for both schema version and API version, but OFF API itself appears to have its own logic for determining which schema version to actually use in the response. This is likely internal API behaviour that we can't control.
We will need more discussions on it as of now, if I'm not wrong in my logic. Please drop down your thoughts about this so that we can find out an approach to solve this.

VaiTon:
@swastikCommits see the linked MD:

> When a client makes a request with a specific API version, we do our best to convert the response to the corresponding product schema version.

Issue #1, opened 2023-07-05
Implement a real login mechanism


VaiTon:
### Depends on
- [x] https://github.com/openfoodfacts/openfoodfacts-server/issues/1596

teolemon:
- Almost there #65

VaiTon:
We still need a way to not send user and passw every time we make a change. Otherwise we have to keep them somewhere (e.g. in LocalStorage)

teolemon:
https://auth.openfoodfacts.org/
https://github.com/openfoodfacts/openfoodfacts-auth

swastikCommits:
The application is storing username and password directly in localStorage through the persisted store. This is a significant security vulnerability.
So @teolemon & @VaiTon  do you want the users to re-direct the users to Keycloak for the auth?

VaiTon:
> The application is storing username and password directly in localStorage through the persisted store. This is a significant security vulnerability.

Exactly. That's why I highlighted it in the UI

![Image](https://github.com/user-attachments/assets/e80d765a-2f18-443b-a79b-2c4777aad22b)

> So @teolemon & @VaiTon do you want the users to re-direct the users to Keycloak for the auth?

We have to wait for the OAuth endpoints to be available. We can't do nothing till that moment.

PR #550, opened 2025-06-18
feat: Implemented real login mechanism 

heymitali:
## Description
This PR implements a secure login and logout flow using Keycloak with PKCE in the OpenFoodFacts Explorer frontend. The flow includes:

 Login:
- Initiates PKCE authentication on login button click.
- Redirects users to Keycloak for authentication.
- Handles the authorization code callback, exchanges it for access, refresh, and ID tokens.
- Stores tokens securely in cookies and updates the user login state.

Logout:
- Initiates logout by redirecting to Keycloak with the ID token.
- Handles the post-logout callback, clears all authentication cookies, and updates the user login state.

State Management:
- Uses a Svelte store userLoginState to track authentication status.
- Checks for valid tokens on app load and refreshes tokens if needed.
## Fixes # (issue)
https://github.com/openfoodfacts/openfoodfacts-explorer/issues/1

## Part of:
https://github.com/orgs/openfoodfacts/projects/157/views/1?pane=issue&itemId=115144718&issue=openfoodfacts%7Copenfoodfacts-explorer%7C524

## Screenshots:

| Home Page Auth Flow | Login Flow | Logout Flow |
|---------------------|------------|-------------|
| Home Page Auth Flow | Login Flow | Logout Flow |

https://github.com/user-attachments/assets/3d723c1a-1e91-471b-9067-99b8d61bc4aa
## Checklist:

**Author Self-Review:**

**LLM Usage Disclosure:**

**Triggering Code Review:**


src/lib/const.ts
```diff
@@ -17,3 +19,15 @@ export const PRODUCT_IMAGE_URL = (path: string) => `${IMAGE_HOST}/images/product
 export const PRODUCT_STATUS = {
 	EMPTY: 'empty'
 };
+
+// Auth/PKCE constants (hardcoded, not from env)
```
VaiTon, line 23: Why not from env?

src/lib/stores/pkceLoginStore.ts
```diff
@@ -0,0 +1,66 @@
+import { AUTH_BASE_URL, AUTH_PKCE_ID, OFF_EXP_BASE_URL } from '$lib/const';
+
+export const keycloak = `${AUTH_BASE_URL}/realms/open-products-facts`;
```
VaiTon, line 3: The realm name should be another env var

src/lib/stores/pkceLoginStore.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export async function getAccessToken(useRefreshToken: boolean = false) {
+	const verifier = localStorage.getItem('verifier');
+	const urlParams = new URLSearchParams(window.location.search);
+	const code = urlParams.get('code');
+
+	const body = new URLSearchParams({
+		code: code ?? '',
+		grant_type: 'authorization_code',
+		redirect_uri: `${OFF_EXP_BASE_URL}/login_callback`,
```
VaiTon, line 22: This should be extracted to a const

src/lib/stores/pkceLoginStore.ts
```diff
@@ -0,0 +1,66 @@
...
+		client_id: pkceClientId,
+		code_verifier: verifier ?? ''
+	});
+
+	if (useRefreshToken) {
+		const refreshToken = document.cookie
+			.split('; ')
```
VaiTon, line 29: Does this flow saves refreshTokens as cookie with the HttpOnly set to false? That's not the best of security

src/lib/stores/pkceLoginStore.ts
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	const response = await fetch(`${keycloak}/protocol/openid-connect/token`, {
+		method: 'POST',
+		body: body,
+		headers: new Headers()
+	});
+
+	const jwt = await response.json();
+	console.log('JWT response:', jwt);
```
VaiTon, line 48: Don't log the JWT response, it's sensitive

src/routes/login/+page.svelte
```diff
@@ -0,0 +1,45 @@
...
+		);
+
+		const nonce = crypto.getRandomValues(new BigUint64Array(1))[0];
+		const lang = 'en';
+
+		const redirectUri = `${OFF_EXP_BASE_URL}/login_callback`;
+
+		const url = `${keycloak}/protocol/openid-connect/auth?response_type=code&client_id=${pkceClientId}&redirect_uri=${encodeURIComponent(redirectUri)}&scope=openid+profile+offline_access&state=${nonce}&ui_locales=${lang}&code_challenge=${codeChallenge}&code_challenge_method=S256`;
+		window.location.href = url;
+	}
```
VaiTon, line 38: Extract to a .ts file

VaiTon (commented):
I'm not really sure about not using a backend at all, because we are exposing tokens one way or the other via cookies or via localstorage. Maybe we could use the backend features of SvelteKit. But then also, how do we use this access_code to query the off api? That's not clear at all to me. @john-gom what do you think?

VaiTon:
Maybe https://auth0.com/blog/the-backend-for-frontend-pattern-bff/ ?

VaiTon:
Also maybe https://authjs.dev/getting-started/authentication/oauth

john-gom:
> I'm not really sure about not using a backend at all, because we are exposing tokens one way or the other via cookies or via localstorage. Maybe we could use the backend features of SvelteKit. But then also, how do we use this access_code to query the off api? That's not clear at all to me. @john-gom what do you think?

We had similar discussions in my previous company. Storing the access and refresh token in the browser can be seen as a security risk, although it is fairly low given the nature of our application (what benefit is there to a malicious actor to contribute rogue product updates on a user's behalf?).

I think the most secure way to do it if we really want to go there is to use secure Http Only cookies. So the access token would only be obtained at the backend which returns the secure cookie to the browser (potentially just the encrypted token). However, this also means that all API calls would have to go via the backend too which would add latency and additional server traffic.

The other way we might be able to do it is to somehow get Product Opener to do the authentication with Keycloak and store a secure cookie in its own domain (openfoodfacts.org) so that this is automatically passed in when we call APIs.

Personally I'm not sure it is worth the effort. Using an access token directly is kind of expected, despite the security concerns. We could add refresh token rotation (can't use the same refresh token twice) if we want to add a bit more security

src/lib/stores/pkceLoginStore.ts
VaiTon, line 3: Also, is `open-products-facts` the right realm?

VaiTon, line 3: Also, this should be in `consts.ts`

src/lib/stores/pkceLoginStore.ts
```diff
@@ -0,0 +1,66 @@
+import { AUTH_BASE_URL, AUTH_PKCE_ID, OFF_EXP_BASE_URL } from '$lib/const';
+
+export const keycloak = `${AUTH_BASE_URL}/realms/open-products-facts`;
+export const pkceClientId = AUTH_PKCE_ID;
+export const accountUrl = `${keycloak}/account/#/`;
```
VaiTon, line 5: To `consts.ts`

src/lib/stores/pkceLoginStore.ts
```diff
@@ -0,0 +1,66 @@
...
+	const body = new URLSearchParams({
+		code: code ?? '',
+		grant_type: 'authorization_code',
+		redirect_uri: `${OFF_EXP_BASE_URL}/login_callback`,
+		client_id: pkceClientId,
+		code_verifier: verifier ?? ''
+	});
+
+	if (useRefreshToken) {
+		const refreshToken = document.cookie
```
VaiTon, line 28: Create a file `cookies.ts` to handle everything related to cookies

src/lib/stores/pkceLoginStore.ts
```diff
@@ -0,0 +1,66 @@
...
+export function saveAuthTokens(jwt: AuthTokens) {
+	// Save accessToken and refreshToken in cookies (expires in jwt.expires_in seconds)
+	const expires = new Date(Date.now() + jwt.expires_in * 1000).toUTCString();
+	document.cookie = `userAccessToken=${jwt.access_token}; expires=${expires}; path=/; secure; samesite=strict`;
+
+	// if a refresh token is provided, set it in cookies
+	if (jwt.refresh_token) {
+		// Set refresh token cookie with a longer expiry if available (e.g., 30 days)
+		const refreshExpires = new Date(Date.now() + 1 * 24 * 60 * 60 * 1000).toUTCString();
+		document.cookie = `userRefreshToken=${jwt.refresh_token}; expires=${refreshExpires}; path=/; secure; samesite=strict`;
```
VaiTon, line 61: Use a function of `cookies.ts`

src/routes/+layout.svelte
```diff
@@ -33,6 +35,33 @@
 	onMount(() => {
 		// only inject the script on the client side
 		injectSpeedInsights();
+
+		// if the cookie userAccessToken is set, set the userLoginState to true
+		const cookies = document.cookie.split('; ');
+		const userAccessToken = cookies.find((cookie) => cookie.startsWith('userAccessToken='));
+		const userRefreshToken = cookies.find((cookie) => cookie.startsWith('userRefreshToken='));
```
VaiTon, line 42: To `cookies.ts`

src/routes/login/+page.svelte
VaiTon, line 1: We don't really need a view, just a function and a callback when the user clicks on the button.

src/routes/login_callback/+page.svelte
VaiTon, line 1: I think we can handle this in +page.ts, without having a view, I'm not sure though

src/routes/logout/+page.svelte
VaiTon, line 1: Same here

src/routes/logout_callback/+page.svelte
VaiTon, line 1: Same here

src/lib/const.ts
```diff
@@ -17,3 +19,15 @@ export const PRODUCT_IMAGE_URL = (path: string) => `${IMAGE_HOST}/images/product
...
+export const AUTH_BASE_URL_PROD = 'https://auth.openfoodfacts.org';
+export const AUTH_BASE_URL_DEV = 'http://auth.openfoodfacts.localhost:5600';
+export const AUTH_PKCE_ID_DEV = 'OFF_EXP_DEV';
```
john-gom, line 26: For dev you can use the pre-built test_public_client now

src/lib/stores/pkceLoginStore.ts
john-gom: Yes, and note this will be changing to "openfoodfacts" soon

src/lib/stores/pkceLoginStore.ts
john-gom: I'm not sure why you are saving the token in a cookie. LocalStorage is just as secure as a cookie that is accessed via JavaScript

heymitali: I was considering using cookies for token storage because they’re automatically included in HTTP requests to the backend. This removes the need to manually attach tokens to each API call from the frontend.
What do you think @VaiTon @john-gom ?

src/lib/const.ts
```diff
@@ -20,14 +20,17 @@ export const PRODUCT_STATUS = {
...
+export const OFF_EXP_BASE_URL_PROD = import.meta.env.VITE_OFF_EXP_BASE_URL_PROD;
+export const OFF_EXP_BASE_URL_DEV = import.meta.env.VITE_OFF_EXP_BASE_URL_DEV;
 
 export const AUTH_BASE_URL = dev ? AUTH_BASE_URL_DEV : AUTH_BASE_URL_PROD;
 export const AUTH_PKCE_ID = dev ? AUTH_PKCE_ID_DEV : AUTH_PKCE_ID_PROD;
 export const OFF_EXP_BASE_URL = dev ? OFF_EXP_BASE_URL_DEV : OFF_EXP_BASE_URL_PROD;
+export const KEYCLOAK_REALM = import.meta.env.VITE_KEYCLOAK_REALM;
+export const KEYCLOAK_URL = `${AUTH_BASE_URL}/realms/${KEYCLOAK_REALM}`;
+export const ACCOUNT_URL = `${KEYCLOAK_URL}/account/#/`;
+export const LOGIN_CALLBACK_URL = `${OFF_EXP_BASE_URL}/login_callback`;
```
VaiTon: We don't need two set of variables, just use one and then we'll change the value depending on the deployment target.

Also, use https://svelte.dev/docs/kit/$env-static-public instead of `import.meta.env`.

src/lib/stores/pkceLoginStore.ts
heymitali: ok. done

src/routes/login_callback/+page.svelte
heymitali: It is not working. I tried that. I needs a +page.svlete.

src/lib/const.ts
heymitali: ok

src/routes/login_callback/+page.svelte
VaiTon: Why wouldn't it work? It is just a redirect. Any particular error message?

src/lib/const.ts
heymitali: done

src/lib/stores/pkceLoginStore.ts
```diff
@@ -7,6 +8,10 @@ export type AuthTokens = {
 	id_token: string;
 };
 
+export const userAccessToken = persisted('userAccessToken', '');
+export const userRefreshToken = persisted('userRefreshToken', '');
+export const userIdToken = persisted('userIdToken', '');
```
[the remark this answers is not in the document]
heymitali: ok

src/lib/stores/pkceLoginStore.ts
```diff
@@ -0,0 +1,57 @@
...
+		}
+
+		body.set('grant_type', 'refresh_token');
+		body.set('refresh_token', refreshToken);
+	}
+
+	const response = await fetch(`${KEYCLOAK_URL}/protocol/openid-connect/token`, {
+		method: 'POST',
+		body: body,
+		headers: new Headers()
```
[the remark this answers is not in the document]
heymitali: I checked and it is not required, so removing it.

src/routes/login/+page.svelte
```diff
@@ -0,0 +1,44 @@
...
+				await crypto.subtle.digest({ name: 'SHA-256' }, new TextEncoder().encode(verifier))
+			)
+		);
+
+		const nonce = crypto.getRandomValues(new BigUint64Array(1))[0];
+		const lang = 'en';
+
+		const redirectUri = `${OFF_EXP_BASE_URL}/login_callback`;
+
+		const url = `${KEYCLOAK_URL}/protocol/openid-connect/auth?response_type=code&client_id=${AUTH_PKCE_ID}&redirect_uri=${encodeURIComponent(redirectUri)}&scope=openid+profile+offline_access&state=${nonce}&ui_locales=${lang}&code_challenge=${codeChallenge}&code_challenge_method=S256`;
```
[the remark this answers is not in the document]
heymitali: ok

heymitali:
@VaiTon , I have removed the code to refetch the accessToken using refreshToken. As mentioned on the off-explorer channel on slack, we will refetch the accessToken for each of the api call, can we pick this task in a different PR?

src/routes/login/+page.svelte
heymitali: done

src/routes/+layout.svelte
```diff
@@ -2,15 +2,18 @@
...
-
 	import { injectSpeedInsights } from '@vercel/speed-insights/sveltekit';
-
 	import Logo from '$lib/ui/Logo.svelte';
 	import Navbar from '$lib/ui/Navbar.svelte';
 	import Footer from '$lib/ui/Footer.svelte';
 	import NutritionCalculator from '$lib/ui/NutritionCalculator.svelte';
+	import '../app.css';
+	import 'leaflet/dist/leaflet.css';
+	import '@fontsource-variable/plus-jakarta-sans';
```
VaiTon, line 12: Why are these 3 lines new?

src/lib/stores/pkceLoginStore.ts
```diff
@@ -0,0 +1,62 @@
...
+		redirect_uri: LOGIN_CALLBACK_URL,
+		client_id: AUTH_PKCE_ID,
+		code_verifier: verifier ?? ''
+	});
+
+	if (useRefreshToken) {
+		let refreshToken = '';
+		userAuthTokens.subscribe((tokens) => {
+			refreshToken = tokens.refreshToken;
+		})();
```
VaiTon, line 33: Use https://svelte.dev/docs/svelte/svelte-store#get here

src/lib/stores/userStore.ts
```diff
@@ -0,0 +1,4 @@
+import { writable } from 'svelte/store';
+
+// Holds whether the user is logged in
+export const userLoginState = writable(false);
```
VaiTon, line 4: We don't need this, make the token store nullable and if null then the user is not logged in

src/routes/login/+page.ts
```diff
@@ -0,0 +1,42 @@
+import { OFF_EXP_BASE_URL, KEYCLOAK_URL, AUTH_PKCE_ID } from '$lib/const';
+
+/**
+ * Encodes a Uint8Array to a base64 URL-safe string.
+ * @param {Uint8Array} bytes - The bytes to encode.
+ * @returns {string} The base64 URL-safe encoded string.
+ */
+function base64URLEncode(bytes: Uint8Array): string {
```
VaiTon, line 8: We don't need, just use `btoa`

https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa

src/routes/login_callback/+page.svelte
VaiTon: Then let's go with it

src/routes/logout_callback/+page.svelte
```diff
@@ -0,0 +1,27 @@
...
+	import { userLoginState } from '$lib/stores/userStore';
+	import { userAuthTokens } from '$lib/stores/pkceLoginStore';
+
+	onMount(() => {
+		// delete all the cookies
+		document.cookie.split(';').forEach((cookie) => {
+			document.cookie = cookie
+				.replace(/^ +/, '')
+				.replace(/=.*/, '=;expires=' + new Date().toUTCString() + ';path=/;secure;samesite=strict');
+		});
```
VaiTon, line 14: This doesn't do anything I guess

src/routes/login_callback/+page.svelte
VaiTon: We should also check that the state is the same, do we do this?

heymitali: which state are we talking about?

src/lib/stores/pkceLoginStore.ts
```diff
@@ -0,0 +1,62 @@
...
+	idToken: string;
+};
+
+export const userAuthTokens = persisted<AuthTokens>('userAuthTokens', {
+	accessToken: '',
+	refreshToken: '',
+	idToken: ''
+});
+
+export async function getAccessToken(useRefreshToken: boolean = false) {
```
VaiTon, line 17: I would split this function into two, the first is called when we're doing the PKCE flow to obtain the token from the verifier, the second is used exclusively to refresh the access_token given a refresh_token

heymitali:
@VaiTon  I have added the checks that match the state parameters

heymitali:
> @heymitali I would really prefer if we had +page.ts instead of +page.svelte, as we don't need them to just redirect the user.

Okay, I will try to do that.

heymitali:
@VaiTon  I have replaced the +page.svelte files with the +page.ts. Please check.

src/routes/+layout.svelte
```diff
@@ -46,7 +46,7 @@
 	// Set up a subscription to the auth tokens to keep login state in sync
 	$effect(() => {
 		const tokens = $userAuthTokens;
-		if (tokens.accessToken) {
+		if (tokens && tokens.accessToken) {
```
VaiTon, line 51:
```suggestion
		if (tokens?.accessToken ?? false) {
```

src/routes/login_callback/+page.ts
```diff
@@ -13,10 +15,22 @@ export const load: PageLoad = async ({ url }) => {
...
+		console.error('Missing authorization code in URL');
+		throw new Error('Authentication failed: Missing authorization code');
+	}
+
 	try {
-		const jwt = await getAccessToken();
+		const jwt = await pkceTokenExchange(verifier, code);
 		saveAuthTokens(jwt);
-		window.location.href = '/';
+		await goto('/');
```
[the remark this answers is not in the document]
heymitali: I mentioned it here: https://github.com/openfoodfacts/openfoodfacts-explorer/pull/550#discussion_r2192709958

PR #751, opened 2025-09-15
feat: add landing page with project overview and navigation options

### LINKED ISSUE #626: Implement a landing page for never seen before and unlogged users
Allow showing a custom page when there's no "already visited once" cookie on the device. It can be as a temporary measure a duplicate of the discover page

VaiTon:
Closes #626


teolemon:
In the app we have a "Reshow onboarding" button in the settings. @VaiTon 
The typical users will be basically people making demos of Open Food Facts. 
But if we make the landing page more substantial than on the classic version, people will want to go back to it (and additional explicit route perhaps ?)

PR #799, opened 2025-09-26
feat: add food logging data viewer

VaiTon:
To test go to <https://explorer-git-flda-openfoodfacts.vercel.app/flda>


john-gom:
Hi @VaiTon this is great to see. I'm going to be on holiday until the 10th October but I will take a look as soon as I can.

src/routes/flda/parsing.ts
```diff
@@ -0,0 +1,35 @@
+export function parseCSV(text: string): Record<string, string>[] {
+	const lines = text.split(/\r?\n/).filter(Boolean);
```
john-gom: This won't handle a newline inside a quoted value. Probably best to use a standard CSV parsing library

john-gom (commented):
It's good to see something here but I'm not sure what it is doing other than showing the data the user could already look at in a spreadsheet.

I'm wondering if an idea might be to transform the data into the user's chosen format, so take a CSV with the metadata for that CSV (ideally in a zip) but have a separate User metadata file which is used to display the contents of the CSV but in the user's own locale. So, the CSV might have a column called "Repas" and the viewer could identify that this is the "meal" column from the CSV metadata and then transform this to user's preferred column heading for meal, e.g. "Pasto"

Obviously anything we do is a bit experimental at this stage so if you have other ideas then please feel free to try more things out. FYI I'm starting work on the repo to hold the specification here: https://github.com/openfoodfacts/food-logging/tree/main

PR #873, opened 2025-12-31
docs: add first-time setup notes for dev server

parinaya05:
Adds helpful notes for first-time contributors to understand
what to expect during the first `pnpm run dev` and tsconfig warnings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,14 @@ First, install dependencies:
 pnpm install
 ```
 
+### First-time Setup Notes
+
+- The first `pnpm run dev` may take 1–2 minutes. This is normal.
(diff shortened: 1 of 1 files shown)


parinaya05:
Hi VaiTon,
Thanks for pointing this out.
I personally didn’t face any major issues with the first-run messages. While going through the project as a first-time contributor, I just felt a bit unsure initially about what to expect during the first run, and I thought a small note might help beginners feel more confident.
That said, I understand your point — if the current messages are already clear enough, this addition may not be necessary. I’m completely fine with removing or updating this based on your suggestion.
Thanks for the review and feedback.
Best regards,
Parinaya

________________________________

PR #880, opened 2026-01-17
feat(compare): implement product comparison feature

VaiTon:
## Description

Add a compare mode, a button on product pages to add a product to compare mode and a button on the global layout to go to the compare mode.

The layout is not really final, just a trial.

Closes #369

### DIFF
--- a/src/lib/api/product.ts
+++ b/src/lib/api/product.ts
@@ -397,7 +397,8 @@ const REDUCED_FIELDS = [
 	'ecoscore_grade',
 	'nova_group',
 	'product_type',
-	'nutriments'
+	'nutriments',
+	'additives_n'
 ] as const;
 
 export type ProductReduced = Pick<Product, (typeof REDUCED_FIELDS)[number]>;


teolemon (approved):
Really great start @VaiTon 👍  Here's a list of all papercuts 
- [ ] Number of additives
- [ ] Nova groups >> Ultra-processing level (with the explicit name of the group)
- [ ] Turn the Eco-Score logo into the Green Score logo (with A+ and F)
- [x] Wrap product name names instead of overlap
- [x] Harmonized miniature size using the same trick as in the mobile app
- [x] Being able to reorder
- [ ] Being able to right click a product in a list to add it to the comparison  (intercepting browser menu ??)
- [ ] Add product from compare mode
- [ ] Change compare icon ?
- [ ] Wrap fruit/veg/nuts
- [ ] Add a button to view ingredients in a modal (to avoid clutter)
- [x] Saturated Fat >> Saturated Fats
- [x] Energy >> Energy (kcal)
- [ ] Optimize for social sharing (On-the fly preview image ? The ability to set a title like "Pierre's comparison" or "The epic biscuit match")

PR #938, opened 2026-02-21
fix(i18n): respect stored user locale over browser language

iMeet07:
## Description

This PR fixes an issue where the application always reset the locale to the browser language on refresh, ignoring the user's previously selected language stored in preferences.

### Changes:
- Normalized stored language codes (`it` → `it-IT`, `en` → `en-US`)
- Updated layout logic to prioritize stored user preference over `window.navigator.language`
- Ensured correct fallback behavior

Now the selected language persists correctly after refresh.

Fixes unexpected locale override on reload.

---

## Checklist:

**Author Self-Review:**

**LLM Usage Disclosure:**

- [x] I used AI assistance for debugging and reasoning but reviewed and validated the changes manually.

### DIFF
--- a/src/routes/+layout.ts
+++ b/src/routes/+layout.ts
@@ -1,12 +1,18 @@
 import { locale, waitLocale } from '$lib/i18n';
 import { browser } from '$app/environment';
 import type { LayoutLoad } from './$types';
-
 export const load: LayoutLoad = async () => {
 	if (browser) {
-		locale.set(window.navigator.language);
-		console.log('i18n: Set locale to', window.navigator.language);
+		const stored = JSON.parse(localStorage.getItem('preferences') || '{}')?.lang;
+
+		const normalized = stored === 'it' ? 'it-IT' : stored === 'en' ? 'en-US' : stored;
+
+		const finalLocale = normalized || window.navigator.language || 'en-US';
+
+		locale.set(finalLocale);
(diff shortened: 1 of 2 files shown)


src/routes/+layout.ts
```diff
@@ -1,12 +1,18 @@
...
-
 export const load: LayoutLoad = async () => {
 	if (browser) {
-		locale.set(window.navigator.language);
-		console.log('i18n: Set locale to', window.navigator.language);
+		const stored = JSON.parse(localStorage.getItem('preferences') || '{}')?.lang;
+
+		const normalized = stored === 'it' ? 'it-IT' : stored === 'en' ? 'en-US' : stored;
+
+		const finalLocale = normalized || window.navigator.language || 'en-US';
```
VaiTon: We already do this in getBrowserLocale, which is called in getLocale. What does this fix?

iMeet07: Thanks for the feedback.
The issue I observed was that when preferences.lang stored a short code like "it", it didn’t match the registered locales ("it-IT"), causing fallback to "en-US" on refresh.
Instead of handling this in +layout.ts, I will move the normalization into getBrowserLocale() to avoid duplication and keep locale resolution centralized.

VaiTon: Thank you! Let me know when the PR is ready :)

PR #939, opened 2026-02-21
fix: use globalThis.fetch for SSR compatibility and ensure string typing for config values

Pankajyadav919:
###Description

This PR improves type safety and environment compatibility in the API layer.

###Changes
Replaced typeof window.fetch with typeof globalThis.fetch in:
   createRobotoffApi
   createKeycloakApi

This removes the browser-only dependency and ensures compatibility across:
   Browser
   Node.js (v18+)
   SSR environments (e.g., SvelteKit)
   Testing environments

Ensured configuration values (ROBOTOFF_URL, KEYCLOAK_URL, OAUTH_CLIENT_ID) are treated as string to satisfy TypeScript requirements when used in APIs such as new URL().

###Motivation
window.fetch is not available in SSR or non-browser environments.
TypeScript infers environment variables as string | undefined, which causes type errors when passed to APIs expecting string.
Explicit string typing ensures the compiler recognizes these values as valid strings.
This PR introduces no runtime behavior changes and focuses purely on improving typing and portability.
Checklist:
 ✅I have performed a self-review of my own code.
✅ I understand the changes I'm proposing and why they are needed.
✅ My changes generate no new warnings or errors (linting, console).
✅ I have made corresponding changes to the documentation (if applicable).

LLM Usage Disclosure:
✅I have reviewed the generated code/text to ensure its accuracy, security, and relevance to the project's context and licensing.

### DIFF
--- a/src/lib/api.ts
+++ b/src/lib/api.ts
@@ -11,14 +11,17 @@ export * from './api/nutriments';
 
 export * from './api/knowledgepanels';
 
-export function createRobotoffApi(fetch: typeof window.fetch) {
-	const { fetch: wrappedFetch, url } = wrapFetchWithCredentials(fetch, new URL(ROBOTOFF_URL));
+export function createRobotoffApi(fetch: typeof globalThis.fetch) {
+	const { fetch: wrappedFetch, url } = wrapFetchWithCredentials(
+		fetch,
+		new URL(ROBOTOFF_URL as string)
+	);
 	return new Robotoff(wrappedFetch, { baseUrl: url.toString() });
 }
 
-export function createKeycloakApi(fetch: typeof window.fetch, url: URL) {
-	const keycloakUrl = KEYCLOAK_URL;
-	const clientId = OAUTH_CLIENT_ID;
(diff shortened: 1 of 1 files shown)


VaiTon:
> Replaced typeof window.fetch with typeof globalThis.fetch in:
> createRobotoffApi
> createKeycloakApi

This only impact types, which are erased during TS -> JS transpilation, so that's not really gonna do anything

> Ensured configuration values (ROBOTOFF_URL, KEYCLOAK_URL, OAUTH_CLIENT_ID) are treated as string to satisfy TypeScript requirements when used in APIs such as new URL().

Does this fix any warning? Using `as` is just a cast, which is unneeded if TS does not complain

Pankajyadav919:
Thanks for the review @VaiTon.

You’re right — changing typeof window.fetch to typeof globalThis.fetch only affects TypeScript types. It doesn’t change the actual JavaScript output, so there’s no runtime impact. I updated it mainly to make it more environment-agnostic, but functionally it doesn’t change behavior.
For the config values, I added as string because TypeScript was treating them as possibly undefined. If TypeScript isn’t complaining in this case, then the cast isn’t really needed.

PR #940, opened 2026-02-21
fix: sending credentials safely

Ramneet04:
Was facing the internal server errors whenever I click on any product.
- The error happened because the API was putting the username and password directly in the URL like `https://user:pass@host`. And it was getting blocked. So instead of sending the request, it throws an error and your server returns a 500. Just removed the credentials from the URL and sending them in headers.


VaiTon:
We don't use credentials in urls anymore. We should just remove them from the example .env file

Ramneet04:
Should I just replace them with with `PUBLIC_OFF_BASE_URL=https://world.openfoodfacts.org`?

PR #969, opened 2026-02-24
feat: implement structured packaging edit

### LINKED ISSUE #764: (Product addition) Add Packaging structured input

swastikCommits:
## Description
It introduces a packaging editor to the Explorer, bringing it in line with the official main OFF website and leveraging the V3 API for structured data.

Key Changes:
- Added a grid to manage multiple packaging parts (e.g., bottle, cap, label) with fields for shape, material, recycling, and weight
- Added a dedicated field for "Recycling instructions and/or packaging information" that adapts to the product's current language
- Added Packaging Image uploading Support
- Made sure its in proper alignment with the SDK and its types

image

Fixes #764 

---

## Checklist:

**Author Self-Review:**

**LLM Usage Disclosure:**

**Triggering Code Review:**

### DIFF
--- a/src/lib/api/product.ts
+++ b/src/lib/api/product.ts
@@ -22,8 +22,6 @@ export type PackagingComponent = {
 	weight_measured?: number;
 };
 
-export { type ProductV3 };
-
 export function createProductsApi(fetch: typeof window.fetch) {
 	const fetchToUse = wrapFetchWithAuth(fetch);
 	const urlToUse = new URL(API_HOST);
@@ -80,12 +78,12 @@ export async function getTaxonomySuggestions(
 	tagtype: string,
 	searchString?: string,
 	limit: number = 25
-): Promise<string[]> {
+) {
 	const off = createProductsApi(fetch);
 	const lc = get(preferences).lang || 'en';
 	const cc = get(preferences).country;
 
-	const { data, error } = await off.apiv3.client.GET('/api/v3/taxonomy_suggestions', {
+	return off.apiv3.client.GET('/api/v3/taxonomy_suggestions', {
 		params: {
 			query: {
 				tagtype,
@@ -96,13 +94,6 @@ export async function getTaxonomySuggestions(
 			}
 		}
 	});
-
-	if (error || !data) {
-		console.warn(`Failed to fetch taxonomy suggestions for ${tagtype}:`, error);
-		return [];
-	}
-
-	return data.suggestions ?? [];
 }
 
 /**
@@ -111,8 +102,10 @@ export async function getTaxonomySuggestions(
  * @param component The raw packaging component
  * @returns A clean object ready for the V3 API
  */
-export function cleanPackagingComponent(component: PackagingComponent): Record<string, unknown> {
-	const cleaned: Record<string, unknown> = {};
+export function cleanPackagingComponent(
+	component: PackagingComponent
+): Record<string, string | number> {
+	const cleaned: Record<string, string | number> = {};
 
 	if (component.number_of_units != null) {
 		cleaned.number_of_units = component.number_of_units;
@@ -120,13 +113,13 @@ export function cleanPackagingComponent(component: PackagingComponent): Record<s
 
 	// For taxonomy fields, send the lc_name (localized name) as a string
 	if (component.shape?.lc_name || component.shape?.id) {
-		cleaned.shape = component.shape.lc_name || component.shape.id;
+		cleaned.shape = component.shape.lc_name || component.shape.id || '';
 	}
 	if (component.material?.lc_name || component.material?.id) {
-		cleaned.material = component.material.lc_name || component.material.id;
+		cleaned.material = component.material.lc_name || component.material.id || '';
 	}
 	if (component.recycling?.lc_name || component.recycling?.id) {
-		cleaned.recycling = component.recycling.lc_name || component.recycling.id;
+		cleaned.recycling = component.recycling.lc_name || component.recycling.id || '';
 	}
 
 	if (component.quantity_per_unit) {
--- a/src/lib/ui/AddProductForm.svelte
+++ b/src/lib/ui/AddProductForm.svelte
@@ -156,14 +156,7 @@
 {:else if currentStep === 4}
 	<NutritionStep bind:product {getNutritionImage} {handleNutrimentInput} />
 {:else if currentStep === 5}
-	<PackagingStep
-		barcode={product.code}
-		lang={product.lang || 'en'}
-		bind:packagings={product.packagings}
-		bind:packagingsComplete={product.packagings_complete}
-		bind:packagingText={product[`packaging_text_${product.lang || 'en'}`]}
-		{getPackagingImage}
-	/>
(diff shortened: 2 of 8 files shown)


src/lib/api/product.ts
```diff
@@ -3,9 +3,18 @@ import { get } from 'svelte/store';
 import type { KnowledgePanel } from './knowledgepanels';
 import type { Nutriments } from './nutriments';
 import { preferences } from '$lib/settings';
-import { type ProductV3, OpenFoodFacts } from '@openfoodfacts/openfoodfacts-nodejs';
+import {
+	type ProductV3,
+	// @ts-expect-error - Packaging types will be available in next SDK version
```
VaiTon: What error are we expecting here? It makes no sense to import something that still doesn't exist.

src/lib/api/product.ts
```diff
@@ -48,6 +57,121 @@ export async function addOrEditProductV2(
...
+			}
+		}
+	});
+
+	if (error || !data) {
+		console.warn(`Failed to fetch taxonomy suggestions for ${tagtype}:`, error);
+		return [];
+	}
+
+	return (data as { suggestions?: string[] }).suggestions ?? [];
```
VaiTon: Why the cast? Is the data type wrong? I would replace it with a `// @ts-expect-error `

src/lib/api/product.ts
```diff
@@ -48,6 +57,121 @@ export async function addOrEditProductV2(
...
 
+/**
+ * Fetch taxonomy suggestions for packaging fields (shapes, materials, recycling)
+ * @param fetch - The fetch function
+ * @param tagtype - The taxonomy type: 'packaging_shapes', 'packaging_materials', or 'packaging_recycling'
+ * @param searchString - Optional search string for autocomplete filtering
+ * @param limit - Maximum number of suggestions (default 25, max 400)
+ * @returns Array of suggestion strings
+ */
+export async function getPackagingTaxonomySuggestions(
```
VaiTon: Isn't there a generic method for this in the SDK? If that's the case, can we make this generic w.r.t. the taxonomy?

src/lib/api/product.ts
```diff
@@ -48,6 +57,121 @@ export async function addOrEditProductV2(
...
+			}
+
+			if (component.quantity_per_unit) {
+				cleaned.quantity_per_unit = component.quantity_per_unit;
+			}
+			if (component.weight_measured != null) {
+				cleaned.weight_measured = component.weight_measured;
+			}
+
+			return cleaned;
```
VaiTon: What this lambda does is not clear at all. Can we extract it to a const / function and document it?

src/lib/ui/edit-product-steps/PackagingStep.svelte
```diff
@@ -0,0 +1,680 @@
...
+				<span class="text-success"
+					>{$_('product.edit.packaging_component.recycling_instructions_examples')}</span
+				>
+			</p>
+		</div>
+	</div>
+</div>
+
+<div class="space-y-4">
+	{#each product.packagings || [] as component, index (index)}
```
VaiTon:
```suggestion
	{#each product.packagings ?? [] as component, index (index)}
```

VaiTon (changes requested):
First round of reviews. Thank you!

src/lib/api/product.ts
swastikCommits: The lambda cleans the packaging component object before sending it to the API ( for example- e.g., pulling out localized strings from taxonomy fields)

Yes, I can extract this into a standalone function so it's much clearer.

src/lib/api/product.ts
swastikCommits: The new SDK `getTaxonomySuggestions` method from PR [#785](https://github.com/openfoodfacts/openfoodfacts-js/pull/785) is not yet available here in explorer's current SDK version yet. 

For now I will maybe rename this local method to a generic `getTaxonomySuggestions` so it can handle any taxonomy. Once we update to newer SDK version in Explorer, we can remove this local wrapper entirely.

VaiTon (changes requested):

src/lib/ui/AddProductForm.svelte
```diff
@@ -152,6 +156,17 @@
 {:else if currentStep === 4}
 	<NutritionStep bind:product {getNutritionImage} {handleNutrimentInput} />
 {:else if currentStep === 5}
+	<PackagingStep
+		barcode={product.code}
+		lang={product.lang || 'en'}
+		bind:packagings={product.packagings}
+		bind:packagingsComplete={product.packagings_complete}
+		bind:packagingText={
+			product[`packaging_text_${product.lang || 'en'}` as keyof Product] as unknown as string
```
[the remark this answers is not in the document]
swastikCommits: indexing `Product` dynamically returns a union of all possible field types, so it had to be explicitly typed to `string` for the `bind:` directive

A cleaner approach imo would be:
```
bind:packagingText={
			(product as Record<string, any>)[`packaging_text_${product.lang || 'en'}`]
		}
```



else we can also do is intersecting the `Product`  type with a generic` Record<string, any>`
this way it product object can act as both the strictly-typed API model and an object capable of holding string keys
```
type Props = {
		product: Product & Record<string, any>;
		// ... other props
};

.....
......

bind:packagingText={product[`packaging_text_${product.lang || 'en'}`]}
```

src/lib/ui/edit-product-steps/PackagingImageUpload.svelte
VaiTon: Instead of re-inventing the wheel, can't we just redirect to image management? Or do like we do in other sections?

src/lib/api/product.ts
```diff
@@ -48,6 +66,130 @@ export async function addOrEditProductV2(
...
+		}
+	});
+
+	if (error || !data) {
+		console.warn(`Failed to fetch taxonomy suggestions for ${tagtype}:`, error);
+		return [];
+	}
+
+	return data.suggestions ?? [];
+}
```
VaiTon: Return at line 88 and then handle errors in the UI layer.

VaiTon (changes requested):
For the free-form edit, we should have language tabs like we have for ingredients.

src/lib/api/product.ts
```diff
@@ -48,6 +66,130 @@ export async function addOrEditProductV2(
...
+
+/**
+ * Cleans a packaging component object before sending it to the API.
+ * This extracts localized strings from taxonomy fields and removes null/empty values.
+ * @param component The raw packaging component
+ * @returns A clean object ready for the V3 API
+ */
+export function cleanPackagingComponent(component: PackagingComponent): Record<string, unknown> {
+	const cleaned: Record<string, unknown> = {};
```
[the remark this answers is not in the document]
swastikCommits: number_of_units, weight_measured are numbers whereas shape, material can be strings
 hence we can use: `Record<string, string | number>` instead

VaiTon (approved):
Good enough for now. We can iterate in main.

PR #977, opened 2026-02-24
fix(robotoff): prevent 404 by correctly configuring PUBLIC_ROBOTOFF_URL

LakshuCodes:
## Description

This PR fixes the Robotoff 404 error on the product page.

### Root Cause

The Robotoff base URL could become `undefined` if `PUBLIC_ROBOTOFF_URL` was not set in the environment.  
This resulted in malformed requests such as:

- `/products/undefined/api/v1/questions/...`
- `/api/v1/api/v1/questions/...`

which caused 404 errors.

### Changes Made

- Updated `const.ts` to ensure `ROBOTOFF_URL` always has a safe fallback (`https://robotoff.openfoodfacts.net`) and never becomes `undefined`.
- Updated `createRobotoffApi` in `api.ts` to ensure the SDK receives a clean base URL (origin only), preventing duplicate `/api/v1` segments.

### Result

Robotoff questions now correctly load from:

Screenshot 2026-02-24 at 7 47 51 PM

### DIFF
--- a/src/lib/api.ts
+++ b/src/lib/api.ts
@@ -12,9 +12,8 @@ export * from './api/nutriments';
 export * from './api/knowledgepanels';
 
 export function createRobotoffApi(fetch: typeof window.fetch) {
-	const cleanBaseUrl = new URL(ROBOTOFF_URL).origin;
-	const { fetch: wrappedFetch } = wrapFetchWithCredentials(fetch, new URL(cleanBaseUrl));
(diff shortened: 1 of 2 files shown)


LakshuCodes:
@VaiTon so should i change the .env.example file only, not this changes that i did right now in this PR.

LakshuCodes:
@VaiTon Thanks a lot for reviewing. And sorry for the repeated tags earlier — still learning the workflow. thanks

PR #1016, opened 2026-02-26
fix(facets): prevent crash when results.products is undefined

### LINKED ISSUE #1014: TypeError: Cannot read properties of undefined (reading 'map')
Sentry Issue: [OPENFOODFACTS-EXPLORER-9F](https://openfoodfacts.sentry.io/issues/7093882258/?referrer=github_integration)

```
TypeError: Cannot read properties of undefined (reading 'map')
  File "../../../../../src/routes/facets/[facet]/[value]/+page.ts", line 38, in productCodes
    const productCodes = (await results).products.map((state) => state.code as string);
```

nagam-surya-narayan:
## Description

Fixes #1014

This PR prevents a runtime crash on facet pages when
`results.products` is undefined.

Previously, the code directly called `.map()` on
`results.products`, which could throw a TypeError if
the API returned an unexpected or empty response.

This update:

- Safely falls back to an empty array when `products` is undefined
- Avoids unnecessary attribute fetching when there are no product codes
- Keeps the existing behavior unchanged for valid responses

This ensures facet pages do not crash even if the API
response is malformed or temporarily empty.

---

## Checklist:

**Author Self-Review:**

**LLM Usage Disclosure:**

### DIFF
--- a/src/routes/facets/[facet]/[value]/+page.ts
+++ b/src/routes/facets/[facet]/[value]/+page.ts
@@ -6,6 +6,7 @@ import {
 	getFacetValue,
 	type FacetSortOption
 } from '$lib/api/facets';
+
 import type { PageLoad } from './$types';
 import { requireInt } from '$lib/utils';
 import { getBulkProductAttributes } from '$lib/api';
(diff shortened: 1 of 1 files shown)


VaiTon (changes requested):

nagam-surya-narayan:
You're right — according to the current TypeScript definition, `products` is not nullable.

Since Sentry reported it as undefined at runtime, this suggests a mismatch between the API response and our type definition.

I'll inspect the return type of `getFacetValue` and verify whether the type needs to be updated instead of adding a fallback here.

VaiTon:
@nagam-surya-narayan are you still working on this? In that case, feel free to open another PR. Closing this as stale.

PR #1018, opened 2026-02-26
fix(ui): polish compare page button styles, text color, and table overflow

nabrahma:
## Description

Several visual issues on the comparison page: joined toggle buttons looked inconsistent, the hint text was unreadable in dark mode, and long nutrient labels overflowed the sticky left column.

### What changed

In `src/routes/compare/+page.svelte`:
- Replaced the `join` button group with individual buttons using `btn-outline` for the active mode and `btn-ghost` for inactive modes.
- Updated the empty state description text from `text-gray-600` to `text-base-content/70` for dark mode readability.

In `src/lib/ui/ComparisonDisplay.svelte`:
- Widened the sticky left column (`w-40` → `w-56`) and added `whitespace-normal leading-tight text-balance` to gracefully wrap long nutrient labels.
- Replaced hardcoded `text-gray-600` with theme-aware `text-base-content/70` for the comparison hint text at the bottom.

### Screenshot

image

---

## Checklist:

**Author Self-Review:**

- [x] I understand the changes I're proposing and why they are needed.

**LLM Usage Disclosure:**

### DIFF
--- a/src/lib/ui/ComparisonDisplay.svelte
+++ b/src/lib/ui/ComparisonDisplay.svelte
@@ -452,7 +452,7 @@
 	<table class="table-zebra table w-full table-fixed">
 		<thead>
 			<tr>
-				<th class="bg-base-100 sticky left-0 z-10 w-56"></th>
+				<th class="bg-base-100 sticky left-0 z-10 w-40"></th>
 				{#each products as product, index (product.code)}
 					<th
 						animate:flip={{ duration: 300 }}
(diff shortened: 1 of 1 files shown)


src/lib/ui/ComparisonDisplay.svelte
```diff
@@ -514,39 +514,39 @@
 		</thead>
 		<tbody>
 			<tr>
-				<td class="bg-base-100 sticky left-0 w-40 font-semibold">Name</td>
+				<td class="bg-base-100 sticky left-0 w-56 font-semibold">Name</td>
```
VaiTon: What if we wrap instead of more width?

nabrahma: I was a bit worried about visual consistency since the other labels sit in one line and only the last long one would wrap.

But yeah wrapping does scale better than bumping the width.

So should I revert the width and warp it instead ?

nabrahma: Ok

nabrahma: @VaiTon Implemented the wrap

PR #1036, opened 2026-02-28
fix: remove loading flash when no prices in Prices.svelte

### LINKED ISSUE #1035: "Loading" flash shown unnecessarily when product has no prices in `Prices.svelte`
# Description

On a product page, the **Prices Map** section briefly flashes **"Loading..."** even when the product has zero prices.

This occurs because the condition tree checks whether `PricesMap`  is `null` **before** checking if `prices.length === 0`.

---

## Steps to Reproduce

1. Open any product page with **no prices submitted**, e.g. `/products/[barcode]`
2. Observe the **Prices Map** section
3. It briefly shows **"Loading..."** before switching to **"No prices found"**

---

## Expected Behavior

If there are **no prices**, the UI should display **"No prices found" immediately**, without flashing **"Loading..."** or triggering the map component import.

puneetnith28:
## Summary

Fixes a UX issue where the **Prices Map** section briefly showed **"Loading..."** before displaying **"No prices found"** for products that have zero prices.

---

## Problem

`PricesMap` is initialized as `null` on mount and only resolves after the lazy import completes (~50–200ms).

The original condition checked `PricesMap == null` first, causing a **"Loading..." flash** before switching to the empty state when `prices.length === 0`.

---

## Fix

Update the condition order to check `prices.length === 0` first.

If there are **no prices**, the UI immediately shows the **empty state** without waiting for the map component to load.

---

## Closes #1035

### DIFF
--- a/src/routes/products/[barcode]/Prices.svelte
+++ b/src/routes/products/[barcode]/Prices.svelte
@@ -26,10 +26,10 @@
 <Card>
 	<h1 class="my-4 text-2xl font-bold sm:text-4xl">Prices Map</h1>
 
-	{#if PricesMap == null}
-		{$_('product.prices.loading')}
-	{:else if prices.length === 0}
+	{#if prices.length === 0}
 		{$_('product.prices.no_prices_found')}
+	{:else if PricesMap == null}
+		{$_('product.prices.loading')}
 	{:else}
 		{#await PricesMap}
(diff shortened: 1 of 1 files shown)


puneetnith28:
Thank you for the reviewing and merging this PR !!
Looking forward to contributing more.

PR #1047, opened 2026-02-28
fix: use semantic span instead of button for decorative elements in Pagination

### LINKED ISSUE #1046: Decorative ellipsis in Pagination uses <button> instead of non-interactive element
## Description
In `Pagination.svelte`, the decorative `...` between page numbers is rendered as a disabled `<button>`.

A `<button>` is meant for user interaction, so using it only for decoration is not correct HTML.

The **current page number** is also rendered as a `<button>` even though clicking it does nothing.

## Impact

- Screen readers announce it as a disabled button, which is confusing
- Keyboard users may tab through these unnecessary elements
- This creates accessibility and semantic HTML issues

## Expected Behavior

- The decorative `...` should use `<span aria-hidden="true">...</span>`
- The current page indicator should also be a `<span>` since it is not clickable

puneetnith28:
## Summary

In `Pagination.svelte`, the decorative `...` and the current page indicator were implemented using `<button>` elements.  
Since these elements are not interactive, this caused accessibility and semantic HTML issues.

---

## Fix

- Replaced disabled `<button>` used for `...` with `<span aria-hidden="true">`
- Replaced the current page `<button>` with `<span aria-current="page">`

---

## Closes #1046

### DIFF
--- a/src/lib/api/search.test.ts
+++ b/src/lib/api/search.test.ts
@@ -1,0 +1,37 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// mutable mock object (values can change inside tests)
+const mockEnv = {
+	PUBLIC_SEARCH_BASE_URL: ''
+};
+
+// mock Svelte public env
+vi.mock('$env/dynamic/public', () => ({
+	get env() {
+		return mockEnv;
+	}
+}));
+
+describe('getSearchBaseUrl', () => {
+	beforeEach(() => {
+		vi.resetModules();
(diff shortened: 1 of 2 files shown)


VaiTon (changes requested):
EDIT: Sorry I got confused.

VaiTon:
Sorry, I already merged https://github.com/openfoodfacts/openfoodfacts-explorer/pull/1058. Didn't see this one. I'll be more careful next time.

PR #1049, opened 2026-02-28
fix: i18n formatTimeSince in DataSources

### LINKED ISSUE #1048: formatTimeSince() in DataSources.svelte hardcoded in English
## Description

The `formatTimeSince()` function in `DataSources.svelte` builds relative time strings (e.g. **"3 days ago"**) by directly concatenating English words instead of using `svelte-i18n`.

Because of this, the time display always appears in English even when the user has selected another language. The rest of the file already uses `$_()` for translations.

## Expected Behavior

The relative time string (e.g. **"3 days ago"**) should use `svelte-i18n` so it appears in the user's selected language.

puneetnith28:
## Summary

The `formatTimeSince()` function in `DataSources.svelte` was building relative time strings (e.g. "3 days ago") directly in English, ignoring the user's selected language.

---

## Problem

While the rest of `DataSources.svelte` correctly uses `svelte-i18n` via `$_()` for user-facing text, `formatTimeSince()` manually constructed English strings with a hardcoded `"ago"` suffix.  
As a result, relative timestamps always appeared in English regardless of the selected language.

---

## Fix

- Updated `formatTimeSince()` to use `$_()` from **svelte-i18n** instead of hardcoded English strings.
- Added missing translation keys in `en-US.json` for relative time units and `just_now`.

This keeps the logic the same while allowing the time display to be translated correctly.

---

## Summary by CodeRabbit

* **New Features**
  * Relative time display for timestamps (e.g., "just now", "2 hours ago").

* **Improvements**
  * Centralized, locale-aware date/time formatting across the app.
  * Safer UI-locale accessor with fallback behavior (publicly available locales).

* **Localization**
  * Added English translation for "just now".

* **Chores**
  * Added dayjs runtime dependency and a new date/time utility.

### DIFF
--- a/src/lib/utils/dayjs.ts
+++ b/src/lib/utils/dayjs.ts
@@ -7,10 +7,12 @@ dayjs.extend(relativeTime);
 dayjs.extend(utc);
 
 const languageCodes = locales.map((locale) => locale.split('-')[0].toLowerCase());
-const dayjsLocaleModules = import.meta.glob('dayjs/locale/*.js', { eager: true });
+const dayjsLocaleModules = import.meta.glob('../../../node_modules/dayjs/locale/*.js', {
+	eager: true
+});
 
 languageCodes.forEach((code) => {
-	const modulePath = `dayjs/locale/${code}.js`;
+	const modulePath = `../../../node_modules/dayjs/locale/${code}.js`;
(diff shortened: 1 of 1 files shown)


VaiTon (changes requested):
Please have a look at https://day.js.org/docs/en/plugin/relative-time

PR #1063, opened 2026-03-01
feat: Add debouncing to search autocomplete to prevent excessive API calls

### LINKED ISSUE #1054: Performance - Missing Debouncing on Search Autocomplete
### Problem

Every keystroke in the search bar triggers an immediate API call to fetch autocomplete suggestions. This causes:
- Excessive API requests (one per character typed)
- Race conditions where older requests can overwrite newer results
- Poor performance on slower connections

### Steps to reproduce:

1. Go to the search page
2. Type a product name quickly (e.g., "coca")
3. Observe multiple network requests in DevTools

### Proposed solution

- API calls should be debounced (300-500ms) to wait for user to stop typing.
- Add a debounce function before calling fetchAutocomplete().

VIDEO:

https://github.com/user-attachments/assets/4c272cba-5758-428f-8cbc-4d7751cb1f36

If this issue looks good to you.Please assign me.

soniya-malviy:
Fixes #1054 

### Description

Every keystroke in the search bar triggered an immediate API call to fetch autocomplete suggestions, causing:
- Excessive API requests
- Race conditions where older requests could overwrite newer results
- Poor performance on slower connections

---

### Solution

Added a 300ms debounce mechanism to delay API calls until the user stops typing.

---

### How It Works

1. User types in search box → oninput fires
2. `debouncedFetchAutocomplete()` is called
3. If a previous timeout exists, it's cancelled (clearTimeout)
4. New timeout is set for 300ms
5. If user types again before 300ms → previous timeout cancelled, new one starts
6. Only when user stops typing for 300ms → `fetchAutocomplete()` actually executes

---

## Checklist:

**Author Self-Review:**

- [✔️ ] I have performed a self-review of my own code.
- [✔️ ] I understand the changes I'm proposing and why they are needed.
- [ ✔️] My changes generate no new warnings or errors (linting, console).
- [ ❌] I have made corresponding changes to the documentation (if applicable).
- [ ✔️] If I **did** use an AI Large Language Model, I have reviewed the generated code/text to ensure its accuracy, security, and relevance to the project's context and licensing.

---

PROOF VIDEO:

https://github.com/user-attachments/assets/26ab36b6-fc5f-4b95-9b62-c34f7ac35ee0

---

`/gemini review`

## Summary by CodeRabbit

* **Performance**
  * Autocomplete now uses debounced requests to reduce unnecessary network calls during rapid typing, improving responsiveness.
* **Reliability**
  * Short queries no longer trigger autocomplete; any in-flight autocomplete requests are canceled when no longer needed.
* **Stability**
  * Pending autocomplete timers and requests are cleaned up when the search component is unmounted to prevent leaks or stray updates.

### DIFF
--- a/src/lib/ui/SearchBar.svelte
+++ b/src/lib/ui/SearchBar.svelte
@@ -3,7 +3,6 @@
 	import { _, getBrowserLocale } from '$lib/i18n';
 	import { onDestroy } from 'svelte';
 
-
 	import IconMdiBarcodeScan from '@iconify-svelte/mdi/barcode-scan';
 
 	let {
@@ -25,7 +24,7 @@
 
 	// debounce for autocomplete
 	let debounceTimeoutId: ReturnType<typeof setTimeout> | undefined;
-	const DEBOUNCE_DELAY_MS = 300;
+	const DEBOUNCE_DELAY_MS = 100;
 
 	// used for aborting previously executing autocomplete requests
 	let autocompleteAbortController: AbortController | null = null;
@@ -68,8 +67,8 @@
 
 	function debouncedFetchAutocomplete(query: string) {
 		if (debounceTimeoutId !== undefined) {
-  			clearTimeout(debounceTimeoutId);
-  			debounceTimeoutId = undefined;
+			clearTimeout(debounceTimeoutId);
+			debounceTimeoutId = undefined;
 		}
 
 		if (query.trim().length < minQueryLength) {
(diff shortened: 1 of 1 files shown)


src/lib/ui/SearchBar.svelte
```diff
@@ -60,6 +64,26 @@
...
+	}
+
+	$effect(() => {
+  		return () => {
+   		  if (debounceTimeoutId !== undefined) {
+		      clearTimeout(debounceTimeoutId);
+		      debounceTimeoutId = undefined;
+	    }
+	  };
+	});
```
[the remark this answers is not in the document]
soniya-malviy: You're right that `debouncedFetchAutocomplete` clears the timeout when the user types again. The `$effect cleanup` serves a different purpose - it clears the timeout when the component unmounts (e.g., user navigates away). Without it, the pending callback could fire after the component is destroyed.

However, I can simplify this to use the more idiomatic `$effect.onCleanup()` pattern in Svelte 5. Want me to update it?

src/lib/ui/SearchBar.svelte
```diff
@@ -60,6 +64,36 @@
...
+			autocompleteList = null;
+			return;  
+   		 }
+
+		debounceTimeoutId = window.setTimeout(() => {
+			fetchAutocomplete(query);
+		}, DEBOUNCE_DELAY_MS);
+	}
+
+	$effect(() => {
```
VaiTon: We can use onDestroy here

soniya-malviy: I'm using $effect cleanup which is the Svelte 5 pattern - is that okay?

PR #1064, opened 2026-03-02
feat: add toast notifications for product edit save feedback

dhiraj-143r:
Adds user-facing feedback when saving a product edit. Previously, users got no visual indication of whether their save succeeded or failed — it just silently logged to console.

Now the  toast system  shows:

* Success toast — "Product saved successfully!" when the API returns OK
* Error toast — "Failed to save product" if API returns failure, or "An error occurred" if there's a network/unexpected error

https://github.com/user-attachments/assets/45355431-7917-46ab-8ac0-a3af7f800333

Part of #125


src/lib/i18n/messages/en-US.json
```diff
@@ -296,6 +296,11 @@
 			"comment": "Comment",
 			"comment_placeholder": "Add a comment to this edit",
 			"save_btn": "Save",
+			"toast": {
+				"save_success": "Product saved successfully!",
+				"save_failed": "Failed to save product. Please try again.",
```
VaiTon: What's the difference between failed and error?

src/routes/products/[barcode]/edit/+page.svelte
```diff
@@ -213,20 +216,31 @@
...
+			});
+		} catch (err) {
+			console.error('Error saving product:', err);
+			toastCtx.error($_('product.edit.toast.save_error'));
 			isSubmitting = false;
-			return;
 		}
-
-		goto('/products/' + product.code, {
-			state: { currentStep: 0 }
```
Comment (VaiTon) [inline src/routes/products/[barcode]/edit/+page.svelte]: Use `finally` to set `isSubmitting` to false

dhiraj-143r:
@VaiTon 
I've updated the code to use the finally block for resetting isSubmitting as requested. I also merged main to resolve the merge conflicts

Issue #1141, opened 2026-03-13
Add reusable CountUp component for animated stats


omarKenawi:
### Summary
Introduce a reusable `CountUp` Svelte component to animate numeric stats when they enter the viewport.  
This can be used for key metrics on the homepage or other sections, such as product counts, contributors, or open data percentages.

### Motivation
Currently, numeric stats are displayed as static numbers. Adding a smooth animation for these values enhances the visual appeal and user engagement on the platform.

### Proposed Solution
- Create a `CountUp.svelte`component.
- Use `Tween` from `svelte/motion` with `cubicOut` easing for smooth number animation.
- Automatically trigger the animation when the element becomes visible using `IntersectionObserver`.
- Accept props for:
  - `value`: the target number to animate to.
  - `duration` (optional): animation duration in milliseconds.
- Format numbers using `Intl.NumberFormat` for readability.

PR #1142, opened 2026-03-13
feat(ui): add reusable CountUp component for animated stats

omarKenawi:
### Summary
This PR addresses issue https://github.com/openfoodfacts/openfoodfacts-explorer/issues/1141 a reusable `CountUp` Svelte component to animate numeric stats when they enter the viewport.  
It improves the UI/UX for key metrics like products count, contributors, or open data percentage on the homepage or other stats sections.

### Features
- Smooth animation using `Tween` from `svelte/motion` with `cubicOut` easing
- Automatically triggers animation when the element becomes visible using `IntersectionObserver`
- Props supported:
  - `value`: target number
  - `duration` (optional): animation duration in milliseconds
- Numbers formatted using `Intl.NumberFormat`
- Fully reusable across multiple cards or sections

### Example Usage
```svelte
<h2 class="text-xl font-bold">
  <CountUp value={data.productCount} />
</h2>

<h2 class="text-xl font-bold">
  <CountUp value={100} suffix="%" />
</h2>
```
### Demo

https://github.com/user-attachments/assets/14a6a4b0-eb7a-4c08-876c-bd4e80f06e08

Fixes https://github.com/openfoodfacts/openfoodfacts-explorer/issues/1141

## Summary by CodeRabbit

* **New Features**
  * Animated number counters on the landing page: product count, contributor count, and completion percentage now animate when scrolled into view.
  * Counters use localized number formatting, support an adjustable animation duration (default 1.5s), and optional suffixes (e.g., "%").
  * Animations begin on first visibility and update smoothly when values change.

### DIFF
--- a/src/lib/ui/CountUp.svelte
+++ b/src/lib/ui/CountUp.svelte
@@ -6,15 +6,17 @@
 	interface Props {
 		value: number;
 		duration?: number;
+		suffix?: string;
 	}
 
-	let { value, duration = 1500 }: Props = $props();
+	let { value, duration = 1500, suffix = '' }: Props = $props();
 
 	const count = new Tween(0, { easing: cubicOut });
 
 	let element: HTMLElement | undefined = $state();
 	let display = $derived(Math.floor(count.current));
 	let hasStarted = false;
+
 	const formatter = new Intl.NumberFormat();
 
 	onMount(() => {
@@ -33,8 +35,14 @@
 
 		return () => observer.disconnect();
 	});
+
+	$effect(() => {
+		if (hasStarted) {
+			count.set(value, { duration });
+		}
+	});
 </script>
 
 <span bind:this={element}>
-	{formatter.format(display)}
+	{formatter.format(display)}{suffix}
 </span>
--- a/src/routes/+page.svelte
+++ b/src/routes/+page.svelte
@@ -161,7 +161,7 @@
 	>
 		<IconMdiLicense class="text-primary mb-4 h-12 w-12" />
 		<h2 class="text-xl font-bold">
-			<CountUp value={100} />%
+			<CountUp value={100} suffix="%" />
 		</h2>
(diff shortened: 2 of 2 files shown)


src/lib/ui/CountUp.svelte
```diff
@@ -0,0 +1,39 @@
...
+		);
+
+		if (element) observer.observe(element);
+
+		return () => observer.disconnect();
+	});
+</script>
+
+<span bind:this={element}>
+	{Intl.NumberFormat().format(display)}
```
VaiTon: Make `Intl.NumberFormat` a component state var. This way we don't need to recreate it for every change.

VaiTon (changes requested):

VaiTon:
Also:

> - 19-34: Consider handling dynamic value prop changes.
> - Enhance the CountUp component with an optional suffix prop for cleaner integration.

from the coderabbit review

src/lib/ui/CountUp.svelte
omarKenawi: Got it, thanks. I've updated it

omarKenawi: @VaiTon 
just an idea we could make the number formatting reactive to the locale so the displayed numbers change based on the user's language

omarKenawi: I noticed that getLocale() returns only 'en' or 'ar', etc., while the NumberFormat constructor expects full locale codes like 'en-US' or 'ar-EG', 'it-IT', etc. I see a few ways to handle this

1- Keep all numbers in English by default, as this is simple and everyone is used to it

2-Map the supported site languages to full locale codes ('en' → 'en-US', 'ar' → 'ar-EG') and pass that to NumberFormat

3-Implement a function that checks the user's selected language and country and falls back to a default if not supported
I think option 2 + a fallback would cover most cases without overcomplicating things

PR #1150, opened 2026-03-15
fix: compare page UI issues

lahari936:
I've successfully pushed the changes to a new branch and they're ready for a Pull Request!

Changes summary:

Added "Images" label to the product image row header.
Standardized alignment by centering content in the score and nutritional value sections.
Updated en-US translations.


VaiTon (changes requested):
Too many changes. PR desc seems AI generated and does not respect template. That's not good.

PR #1151, opened 2026-03-15
fix: potential N+1 API calls when fetching prices

### LINKED ISSUE #1147: Bug: Potential N+1 API calls when fetching prices on the search results page
###  Bug: Possible N+1 API calls when fetching prices on the search results page

**Description**

While reviewing the search results implementation, I noticed that price information for products appears to be fetched individually for each product returned by a search query.

In `src/routes/search/+page.ts`, price requests are executed inside a `Promise.all` loop that calls the Prices API separately for every barcode. If a search returns a full page of results (for example ~24 products), this results in approximately 24 simultaneous API requests.

```ts
const results = await Promise.all(
  barcodes.map(async (code) => ({
    code: code,
    prices: await api.getPrices({ product_code: code })
  }))
);
```

**Steps to Observe**

1. Open the search page.
2. Perform a search that returns multiple products.
3. Open the browser **Developer Tools → Network** tab.
4. Observe multiple requests being sent to the Prices API (one request per product).

**Expected Behavior**

Price data should ideally be fetched in a more efficient way that avoids triggering many simultaneous API requests for a single search page.

**Actual Behavior**

The current implementation sends one request per product barcode, which can result in many concurrent requests when displaying search results.

**Potential Impact**

* Increased network traffic
* Slower loading times on slower connections
* Possible rate-limit pressure on the Prices API
* Reduced performance on mobile devices

**Possible Direction**

If supported by the backend, batching price requests for multiple product codes could reduce the number of API calls. Another option could be deferring price fetching so the main search results render immediately while price data loads progressively.

**File Reference**

`src/routes/search/+page.ts`

student-muskankumari:
## Description

This PR fixes the issue where price data for every product on the search results page was being fetched all at once using "Promise.all". When there are a lot of search results, this causes a huge spike in API requests which can overload the Prices API or trigger rate limiting.

##Solution

To fix this, I added a concurrency limiter that allows only **5 requests to run at a time**:
```js
const PRICES_CONCURRENCY_LIMIT = 5
```

The remaining requests queue up and execute as slots free up, which makes the whole thing a lot more stable.

Fixes #1147 

---

## Checklist:

**Author Self-Review:**

**LLM Usage Disclosure:**

**Triggering Code Review:**

Requesting review from maintainers. Please let me know if any changes or improvements are required.

## Summary by CodeRabbit

* **New Features**
  * Added concurrency-limited execution for price lookups to improve throughput and resource use.
  * Introduced a reusable concurrency utility to run bounded async tasks in original order.

* **Bug Fixes**
  * Empty barcode queries now return immediately with no work.
  * Individual lookup failures are logged and defaulted to zero without aborting the overall fetch.
  * Zero totals are preserved correctly.

### DIFF
--- a/src/lib/utils/promises.ts
+++ b/src/lib/utils/promises.ts
@@ -1,0 +1,47 @@
+/**
+ * Runs an array of async tasks with a bounded number of concurrent executions.
+ *
+ * Unlike `Promise.all`, this does not fire every task simultaneously. Instead it
+ * spins up at most `limit` workers that each pull the next task from a shared
+ * queue, which keeps downstream APIs from being overwhelmed.
+ *
+ * Each task's outcome is reported independently via `PromiseSettledResult`, so
+ * a single rejection never short-circuits the rest of the batch.
+ *
+ * @param tasks  Factory functions that each produce a Promise when invoked.
+ * @param limit  Maximum number of tasks running at the same time. Values that
+ *               are not finite positive numbers are clamped up to `1`.
+ * @returns      One settled result per input task, in the original order.
+ */
+export async function withConcurrencyLimit<T>(
+	tasks: (() => Promise<T>)[],
+	limit: number
+): Promise<PromiseSettledResult<T>[]> {
+	if (tasks.length === 0) return [];
+
+	// Normalise the limit to a safe positive integer so a bad input can never
+	// stall the worker pool.
+	const safeLimit = Number.isFinite(limit) && limit > 0 ? Math.floor(limit) : 1;
+
+	const results: PromiseSettledResult<T>[] = new Array(tasks.length);
+	let nextIndex = 0;
+
+	const worker = async () => {
+		while (true) {
+			const currentIndex = nextIndex++;
+			if (currentIndex >= tasks.length) return;
+
+			try {
+				const value = await tasks[currentIndex]();
+				results[currentIndex] = { status: 'fulfilled', value };
+			} catch (reason) {
+				results[currentIndex] = { status: 'rejected', reason };
+			}
+		}
+	};
+
+	const workerCount = Math.min(safeLimit, tasks.length);
+	await Promise.all(Array.from({ length: workerCount }, worker));
+
+	return results;
+}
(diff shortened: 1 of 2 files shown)


roshinit-a (commented):
Nice work on this! Resolving the N+1 issue for the Prices API is a great catch and will really help with the project's stability.

I have a couple of suggestions to make this even better:

1. **Helper Reusability**: The `withConcurrencyLimit` function you wrote is a fantastic utility. What do you think about moving it to a shared file like `src/lib/utils/promises.ts`? That way, we can reuse it for other API-heavy parts of the app.
2. **Style Tweak**: I noticed some dashed-line separators (`//-------`) in the code. We usually don't use those in this codebase, so it might be worth removing them to keep the file consistent with the rest of the project.
3. **Error Handling**: Really smart use of `PromiseSettledResult`. It’s great that one failed price lookup won't crash the entire search results page now!

Let me know what you think!

student-muskankumari:
Hi @VaiTon @roshinit-a 

Sorry for the long delay — I finally pushed the updates!

I rebased the branch onto the latest main and squashed everything into a single clean commit, so all the old conflicts and clutter should be gone now.

Here's what I addressed:

**@roshinit-a's suggestions:**

- Moved **withConcurrencyLimit** into a shared file at **src/lib/utils/promises.ts** so it can be reused in other parts of the app later
- Removed the **//-----** separator lines to match the rest of the codebase
- Kept the **PromiseSettledResult** error isolation as you suggested

**From the Copilot reviews:**

- Cleaned up the grammar and style in the comments
- Added validation for the **limit** param (safely clamps 0, negative, or NaN to 1)
- Rejected tasks now carry their barcode with them, so the log actually tells you which product failed
- Renamed the confusing **prices** property to **pricesResponse**
- Fixed the edge case where products with 0 prices were being dropped from the map — now any numeric **total** gets stored, including 0
- Ran Prettier so formatting is consistent
- Cleaned up the quotes in the template literal

I also ran it locally against prices.openfoodfacts.net with a chocolate search (24 products). The concurrency stays bounded at 5 across multiple reloads, and the page renders correctly — here's the log output from one of the runs:

    [getPrices] ===== START ===== 24 barcodes, limit=5
    [getPrices] ===== END   ===== took 1987ms, max concurrent = 5

Let me know if anything else needs changing. Thanks a lot for the patience and the thorough reviews!

VaiTon:
@coderabbitai resume

student-muskankumari:
Hi @VaiTon,

Just a gentle follow-up- pushed a small follow-up commit (80c6cd7) addressing CodeRabbit's nit about throwing a plain object. Rejections now throw a proper Error with the original error set as cause and the barcode attached as code. Stack traces are preserved and the consumer contract stays the same.

All checks are green and there are no conflicts. Let me know if you need any further changes!

Thanks again for the thorough reviews.

PR #1165, opened 2026-03-16
fix: improve type safety for OCR API response

devamithra137:
## Description

This PR improves type safety when handling the OCR API response in IngredientsStep.svelte.

Previously, the response from performOCR was force-cast using as OCRResult. Since the API response type is not strictly enforced, this could lead to unsafe assumptions if the returned data structure is unexpected.

This update adds a validation step to ensure that the response exists and is an object before performing the cast. This helps prevent potential runtime errors while keeping the existing functionality unchanged.

Fixes # (issue)

---

## Checklist:

**Author Self-Review:**

- [ x] I have performed a self-review of my own code.
- [x ] My changes generate no new warnings or errors (linting, console).

**LLM Usage Disclosure:**

- [ x] If I **did** use an AI Large Language Model, I have reviewed the generated code/text to ensure its accuracy, security, and relevance to the project's context and licensing.

**Triggering Code Review:**

### DIFF
--- a/src/lib/ui/edit-product-steps/IngredientsStep.svelte
+++ b/src/lib/ui/edit-product-steps/IngredientsStep.svelte
@@ -55,6 +55,11 @@
 				return;
 			}
 
+			if (!tmpData || typeof tmpData !== 'object') {
+				console.warn('OCR returned invalid data:', tmpData);
+				return;
+			}
+
 			const data = tmpData as OCRResult;
 			if (!data || typeof data !== 'object') {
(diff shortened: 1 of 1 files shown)


VaiTon:
> Previously, the response from performOCR was force-cast using as OCRResult. Since the API response type is not strictly enforced, this could lead to unsafe assumptions if the returned data structure is unexpected.

This does not add any check to that, it just checks that it is an object (which would be difficult to not be).

PR #1179, opened 2026-03-17
refactor: parallelize api fetches in product loader

Pavanboga07:
## Description
Hey team! 👋

While I was looking around the codebase to learn how the data fetching works, I noticed something interesting in the product details loader (src/routes/products/[barcode]/+page.ts).

Right now, the loader asks the API for the main product data and awaits the response. Only after that main request completely finishes does it start asking for everything else, like prices, folksonomy tags, and categories.

Because none of those extra requests actually depend on the data returned from the first request (they only need the barcode from the URL), they are just sitting around waiting for no reason! This creates a network waterfall where the total loading time is the time of the product fetch plus the time of the secondary fetches.

I just swapped the order of operations a bit. I moved all the secondary promises up to the top of the function so they fire off immediately. Then, the main product call happens at the exact same time. The loader now fetches everything entirely in parallel and just awaits them all at the bottom before returning.

It's a really tiny change, but in my local testing, this effectively cuts the page load latency in half since we aren't stringing the network requests together sequentially anymore!

## Checklist:

**Author Self-Review:**

**LLM Usage Disclosure:**

## Summary by CodeRabbit

* **Refactor**
  * Product page data loading now runs multiple requests in parallel for faster page readiness.
* **Bug Fixes**
  * Improved error propagation so all parallel fetches settle before failures are reported, reducing inconsistent errors and ensuring connectivity issues return consistent responses.
* **Behavior**
  * Price initialization moved earlier to enable concurrent price processing.
* **New Features**
  * Product pages include richer metadata (tags, taxonomies, origins, brands/stores, default preferences, prices).

### DIFF
--- a/src/routes/products/[barcode]/+page.ts
+++ b/src/routes/products/[barcode]/+page.ts
@@ -109,6 +109,7 @@ export const load: PageLoad = async ({ params, fetch }) => {
 		state = result.data;
 		apiErrorWrapped = result.error;
 	} catch (err) {
+		// Prevent unhandled promise rejections if the main fetch throws unexpectedly
 		void Promise.allSettled([folksonomyTags, folksonomyKeys, pricesResponse, defaultPreferences]);
 		throw err;
 	}
@@ -120,6 +121,7 @@ export const load: PageLoad = async ({ params, fetch }) => {
 	handleProductApiError(apiErrorWrapped);
 
 	if (!state) {
+		// Also settle promises if we are going to throw an error response
 		void Promise.allSettled([folksonomyTags, folksonomyKeys, pricesResponse, defaultPreferences]);
 		throw error(500, {
 			message: 'Unable to connect to Open Food Facts API',
--- a/src/routes/products/[barcode]/+page.ts
+++ b/src/routes/products/[barcode]/+page.ts
@@ -115,9 +115,9 @@ export const load: PageLoad = async ({ params, fetch }) => {
 	}
 
 	if (apiErrorWrapped) {
+		// Settle parallel promises before throwing API errors (404/400/etc.)
 		void Promise.allSettled([folksonomyTags, folksonomyKeys, pricesResponse, defaultPreferences]);
 	}
-
 	handleProductApiError(apiErrorWrapped);
 
 	if (!state) {
--- a/src/routes/products/[barcode]/+page.ts
+++ b/src/routes/products/[barcode]/+page.ts
@@ -52,20 +52,20 @@ function handleProductApiError(apiErrorWrapped: ProductStateResponse | null | un
 	}));
 
 	if (isInvalidFormat) {
-		error(400, {
+		throw error(400, {
 			message: ERR_INVALID_BARCODE,
 			errors: cleanErrors
 		});
 	}
 
 	if (err.result?.id === 'product_not_found') {
-		error(404, {
+		throw error(404, {
 			message: ERR_PRODUCT_NOT_FOUND,
 			errors: cleanErrors
 		});
 	}
 
-	error(500, {
+	throw error(500, {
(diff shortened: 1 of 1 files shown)


Pavanboga07:
Hey @VaiTon

Quick follow-up with numbers + a bit more detail on what I actually did here.

I ran a small before/after benchmark locally (same set of barcodes, multiple runs) and compared the distributions instead of cherry-picking a single load:

Median loader completion time: ~15.8% faster
p95: ~13.9% faster
Mechanically, the change is just about avoiding the network waterfall: kick off the requests that only need the barcode (prices / folksonomy / attribute groups + default prefs) immediately so they overlap with the main product fetch, and only keep the genuinely-dependent stuff (like product-type-aware taxonomy) after we have the product response.

I also tightened the failure paths while I was in there: making sure we throw error(...) so SvelteKit control flow stops cleanly, and settling in-flight promises to avoid unhandled rejections without delaying the thrown response.

VaiTon:
I'm hesitant to consider fetching things in parallel because we already hit some 429, and we should switch to batch APIs.

Pavanboga07:
Hey @VaiTon, that's a fair point and honestly I didn't think about the 429 side of things when I wrote this, so thanks for bringing it up!

Just wanted to mention though, this change doesn't actually send any extra requests. It's the same calls as before, just happening at the same time instead of one after another. And since they're going to different endpoints I didn't think it would make the rate limiting worse, but I get the hesitation if 429s are already a problem.

Also really keen to help out with the batch API migration if that's something that's being planned. Performance work is exactly the kind of thing I want to get more involved in with this project. So if there's a tracking issue or anything I can pick up around that, just let me know!

Happy to close this or put it on hold if it's not the right time for it.

src/routes/products/[barcode]/+page.ts
```diff
@@ -52,20 +52,20 @@ function handleProductApiError(apiErrorWrapped: ProductStateResponse | null | un
 	}));
 
 	if (isInvalidFormat) {
-		error(400, {
+		throw error(400, {
```
Comment (VaiTon) [inline src/routes/products/[barcode]/+page.ts:55]: `error` already throws, no need to throw manually

src/routes/products/[barcode]/+page.ts
```diff
@@ -74,31 +74,70 @@ function handleProductApiError(apiErrorWrapped: ProductStateResponse | null | un
...
+			// @ts-expect-error - This is a temporary workaround until the SDK supports this parameter.
+			knowledge_panels_client: 'web'
+		});
+		state = result.data;
+		apiErrorWrapped = result.error;
+	} catch (err) {
+		// Prevent unhandled promise rejections if the main fetch throws unexpectedly
+		void Promise.allSettled([folksonomyTags, folksonomyKeys, pricesResponse, defaultPreferences]);
+		throw err;
+	}
```
Comment (VaiTon) [inline src/routes/products/[barcode]/+page.ts:115]: I don't think we need this. At least until we see that this is a real leak and that it can happens.

src/routes/products/[barcode]/+page.ts
```diff
@@ -74,31 +74,70 @@ function handleProductApiError(apiErrorWrapped: ProductStateResponse | null | un
...
+
+	if (apiErrorWrapped) {
+		// Settle parallel promises before throwing API errors (404/400/etc.)
+		void Promise.allSettled([folksonomyTags, folksonomyKeys, pricesResponse, defaultPreferences]);
+	}
```
Comment (VaiTon) [inline src/routes/products/[barcode]/+page.ts:120]: Same here

VaiTon (changes requested):

VaiTon:
Just keep the code simple, no need to overengineer with a try-catch here

Pavanboga07:
okay i will keep it simple and try to improve the performance

Issue #986, opened 2026-02-25
Bug: Missing response.ok checks cause unhandled errors in multiple API calls


rohanshrma222:
### What

Multiple API call sites in the codebase are missing response.ok checks before parsing JSON responses, causing unhandled promise rejections when APIs return error responses.
### Steps to reproduce the behavior:
1. Make an API call that returns an error
2. The code attempts to parse JSON without checking if the response was successful
3.  Application crashes or throws unhandled errors

### Expected behavior

All API calls should check response.ok (or handle response.status) before attempting to parse the response body.
### Platform (Desktop, Mobile, Hunger Games)
 - OS: All
 - Platform Web Application

### Part of
- This is a general bug affecting multiple files.

PR #1195, opened 2026-03-18
fix: use correct error state in BarcodeInfo catch handlers

### LINKED ISSUE #1189: Bug: Missing response.ok checks and incorrect error handling in BarcodeInfo.svelte
## Description

While PR #994 resolved the majority of missing `response.ok` checks raised in Issue #986, a few edge cases in `BarcodeInfo.svelte` were missed and still need to be addressed.

### 1. Incorrect error state in catch handlers
The `catch` handlers for `openPricesStatus` and `proOffStatus` currently set the status to `0` (which represents a **loading** state) when an API call fails. This means the user sees a perpetual loading spinner instead of an error message.

The component uses the following status code convention:

| Status Code | Meaning |
|---|---|
| `0` | Loading (request in progress) |
| `1` | Success (data loaded) |
| `-1` | Error (request failed) |

```js
// Current (incorrect)
catch (error) {
    openPricesStatus = 0; // 0 = loading, user never sees the error
}
```

The status should be set to `-1` to indicate an error state:
```js
// Expected (correct)
catch (error) {
    openPricesStatus = -1; // -1 = error, user sees an error message
}
```

### 2. Incorrect timestamp field usage
The code uses `product?.rev` (a revision number / integer counter) as a modification timestamp. It should use `product?.last_modified_t`, which is the actual UNIX timestamp representing the last modification time.

> **Note:** If `last_modified_t` is `null` or `undefined`, the existing fallback behavior should be preserved to avoid breaking the UI. The fix should only replace `rev` with `last_modified_t` where the value is available.

## Expected Behavior
- API call failures should show an error state to the user, not a loading spinner.
- Modification timestamps should use the correct `last_modified_t` field.

## Related
- Issue #986
- PR #994 (merged, fixed most `response.ok` checks but missed these edge cases)

roshinit-a:
Fixes #1189

## Description

This PR fixes incorrect error handling in the `BarcodeInfo.svelte` component. These are edge cases that were missed by PR #994 (which resolved the majority of Issue #986).

### Subsequent Refactor
Refactored the `BarcodeInfo.svelte` component to use Svelte 5 `$derived` Promises and `{#await}` blocks to handle API states instead of relying on manually updated numeric status codes. 

### Key Changes
* **Removed Magic Numbers**: Replaced the `openPricesStatus` and `proOffStatus` integer state variables (which used `-1`/`0`/`200`/`404`) with derived promises.
* **Simplified Reactivity**: Replaced the older `$effect` based fetching with `$derived` promises (`openPricesPromise` and `proOffPromise`). If the `code` prop updates, the promises are now properly re-evaluated automatically.
* **Declarative UI**: Replaced ternary operators in the template with explicit `{#await}` blocks, which correctly distinguish between the pending `…` state, success, and error states (such as `Not found` or `Error`).
* **SSR Safe**: Swapped `window.fetch` with `fetch` to ensure the component does not break during Server-Side Rendering (SSR).

## Summary by CodeRabbit

* **Refactor**
  * Improved internal error handling and state management for product information retrieval, resulting in more consistent error messaging and loading state behavior.

### DIFF
--- a/src/lib/ui/BarcodeInfo.svelte
+++ b/src/lib/ui/BarcodeInfo.svelte
@@ -15,10 +15,12 @@
 			.then((r) => {
 				openPricesStatus = r.data && r.data.items && r.data.items.length > 0 ? 200 : 404;
 			})
-			.catch(() => (openPricesStatus = 0));
+			.catch(() => (openPricesStatus = -1));
 		fetch(`https://pro.openfoodfacts.dev/products/${code}`)
-			.then((r) => (proOffStatus = r.status))
-			.catch(() => (proOffStatus = 0));
+			.then((r) => {
+				proOffStatus = r.ok ? 200 : r.status;
+			})
+			.catch(() => (proOffStatus = -1));
 	});
 </script>
 
@@ -66,9 +68,11 @@
 					>
 						Open Prices {openPricesStatus === null
 							? '…'
-							: openPricesStatus === 200
-								? ''
-								: ' (Not found)'}
+							: openPricesStatus === -1
+								? ' (Error)'
+								: openPricesStatus === 200
+									? ''
+									: ' (Not found)'}
 					</a>
 				</li>
 				<li>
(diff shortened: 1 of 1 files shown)


VaiTon (commented):
This component was a prototype. If we're gonna touch it, we need to rethink it and rethink its state. I don't really like having sentinel values around as 0, 1, ...

VaiTon:
Closing as superseded by #1266

Issue #1244, opened 2026-03-23
Bug: `idToName()` in location.ts returns `undefined` instead of `'Unknown'` when no name tag exists


roshinit-a:
## Description

The `idToName()` function in `src/lib/location.ts` has a type contract of `Promise<string>` and explicitly returns `'Unknown'` when the Overpass API returns an empty `elements[]` array (line 71). However, there is a **missing fallback** for when `elements` is non-empty but the returned OSM node has **none** of the expected name tags.

## Affected File

`src/lib/location.ts` — lines 74–83

## Current (Buggy) Code

```ts
const tags = res.elements[0].tags;
let name = tags.name;
if (name === undefined) {
    name = tags['brand:name'];
}
if (name === undefined) {
    name = tags['brand'];
}
return name; // ← returns `undefined` if ALL three tags are missing!
```

If an OSM element exists but has none of the `name`, `brand:name`, or `brand` tags (e.g., a node with only an `operator` or `ref` tag), all three assignments leave `name` as `undefined` and the function silently returns `undefined` — breaking the `Promise<string>` type signature.

## Steps to Reproduce

1. Call `idToName(fetch, <id>)` with an Overpass node ID whose OSM tags don't include `name`, `brand:name`, or `brand` (e.g., a location tagged only with `operator`, `ref`, or `website`).
2. The function returns `undefined` instead of `'Unknown'`.
3. Any caller that treats the result as a string (e.g., string interpolation, `.toUpperCase()`) will throw a runtime TypeError.

## Expected Behavior

The function should return `'Unknown'` as a safe fallback in all cases where a human-readable name cannot be determined — consistent with the existing fallback on line 71.

## Proposed Fix

Change line 83 from:
```ts
return name;
```
to:
```ts
return name ?? 'Unknown';
```

This is a one-line fix that covers all three missing-tag scenarios while preserving the existing logic.

## Additional Context

- The function is typed as `Promise<string>`, so returning `undefined` is also a TypeScript type violation (masked by implicit `any` propagation in callers).
- No existing issue covers this specific edge case.

PR #1245, opened 2026-03-23
fix(location): change return type to string | undefined for missing location names

roshinit-a:
### Description

`idToName()` in `src/lib/location.ts` had a type contract mismatch where it promised `Promise<string>` but could return `undefined` when an OSM node exists but lacks `name`, `brand:name`, or `brand` tags.

**Solution:** Changed the return type to `Promise<string | undefined>` to accurately represent that location names may not always be available from the Overpass API.

Per maintainer feedback (VaiTon, Mar 29), presentation fallbacks like "Unknown" should be handled at the UI layer, not in the data-fetching layer. This change maintains proper separation of concerns.

### Changes Made
- ✅ Updated return type from `Promise<string>` to `Promise<string | undefined>`
- ✅ Removed hardcoded `'Unknown'` fallback from the data layer
- ✅ Function now truthfully returns `undefined` when no name tags exist
- ✅ UI components can now handle the display logic for missing names

### Screenshot or video

*No visual changes (this is a type-safety and architectural fix).*

* Addresses: [Bug: `idToName()` in location.ts returns `undefined` instead of `'Unknown'` when no name tag exists #1244](https://github.com/openfoodfacts/openfoodfacts-explorer/issues/1244)

---

* ✅ I have performed a self-review of my own code (including running it).
* ✅ I understand the changes I'm proposing and why they are needed.
* ✅ My changes generate no new warnings or errors (linting, console).
* ✅ I have made corresponding changes to the documentation (if applicable).
* ✅ I have addressed maintainer feedback regarding architectural concerns.

* [ ] Generic LLM v0.0.0

### DIFF
--- a/src/lib/const.ts
+++ b/src/lib/const.ts
@@ -1,5 +1,6 @@
 import { resolve } from '$app/paths';
 import { env as publicEnv } from '$env/dynamic/public';
+import { toWebsiteFlavor, WEBSITE_FLAVOR_METADATA } from '$lib/flavor';
 
 const {
 	PUBLIC_ROBOTOFF_URL,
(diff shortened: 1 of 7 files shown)


VaiTon (changes requested):
This layer should not handle missing names. It should be handled at the UI level. If the function return type is incorrect we should fix that.

PR #1247, opened 2026-03-23
fix(service-worker): cache only static and same-origin assets

### LINKED ISSUE #1246: Bug: Service worker caches ALL 200 GET responses, including live API calls
## Description

The service worker in `src/service-worker.ts` is designed to apply **cache-first** strategy only for `https://static.openfoodfacts.org` (line 41). However, the `cache.put()` call (lines 61–63) sits in the general network-first `try` block and runs for **every** successful GET response — including live API calls to `world.openfoodfacts.org`, the search API, Robotoff, Open Prices, etc.

## Affected File

`src/service-worker.ts` — lines 40–63

## Current (Buggy) Code

```ts
// for static assets, we want to cache-first
if (url.origin === STATIC_HOST) {
    const cached = await cache.match(event.request);
    if (cached) {
        console.log(`SW: cache hit for ${event.request.url}`);
        return cached;  // ← cache-first only for static
    }
}

// for everything else, try the network first...
try {
    const response = await fetch(event.request);
    // ...
    if (response.status === 200) {
        cache.put(event.request, response.clone()); // ← BUG: caches ALL 200 GETs, not just static
    }
    return response;
}
```

## Impact

- **Stale product data** — product pages, nutriments, ingredients, and prices can be silently served from a stale cache after the user navigates to a product they visited before.
- **Stale search results** — autocomplete and search-a-licious results can be served from cache.
- **Broken offline fallback logic** — the cache also ends up holding API responses that should never be served as offline fallbacks (e.g., a cached empty search result being returned while offline).

## Expected Behavior

Only responses from `STATIC_HOST` (`https://static.openfoodfacts.org`) should be cached. All other requests (API calls, dynamic routes) should **not** be stored in the cache — they should only fall back to cache if the network is unavailable (the current `catch` branch already handles this correctly for static assets).

## Proposed Fix

Restrict `cache.put()` to only fire when the request origin is `STATIC_HOST`:

```ts
if (response.status === 200 && url.origin === STATIC_HOST) {
    cache.put(event.request, response.clone());
}
```

This is a **one-line change** that aligns the caching write with the cache-read guard already in place on line 41.

roshinit-a:
### Description

The service worker's `cache.put()` call was placed in a general network-first `try` block, causing it to run for **every** successful GET response. This meant live API responses (product data, search results, etc.) were being stored in the cache and could be served as stale data on subsequent visits.

This PR restructures the fetch handler into three explicit origin-based branches with precise caching rules:

```
GET request
│
├── chrome-extension:// → ignored entirely
├── non-GET             → ignored entirely
│
├── origin === STATIC_HOST    → cache-first  (static.openfoodfacts.org images/assets)
│                               cache write: if response.ok || response.type === 'opaque'
│                               wrapped in: event.waitUntil()
│
├── isSameOriginAsset         → network-first with offline fallback
│   (destination === 'script'   cache write: if response.status === 200 only
│    or 'style', or .js/.css     wrapped in: event.waitUntil()
│    pathname fallback)
│
└── everything else           → plain fetch(), NEVER cached
    (API calls to world.openfoodfacts.org, prices API, Robotoff, etc.)
```

### Key changes

- **`STATIC_HOST` branch**: cache-first. Writes to cache if `response.ok || response.type === 'opaque'` to correctly handle opaque cross-origin responses (which have status `0`, not `200`). Cache write wrapped in `event.waitUntil()` to prevent premature service worker termination.

- **`isSameOriginAsset` branch**: network-first, offline fallback. Narrowed from the previous broad `url.origin === self.location.origin` to only match SvelteKit static build artifacts — `request.destination === 'script' | 'style'`, with `.js`/`.css` pathname fallbacks for programmatic `fetch()` calls where `destination` can be `''`. Navigation requests (`destination === 'document'`) and same-origin API routes (`+server.ts`) both fall through to the passthrough branch. Cache write wrapped in `event.waitUntil()`.

- **Everything else**: plain `fetch()` passthrough — no cache reads, no cache writes.

- **`caches.open(CACHE)` is lazy-loaded** inside each branch, not at the top of `respond()`, so external API calls never incur unnecessary cache storage overhead.

### Screenshot or video

_No visual changes — this is a service worker caching logic fix._

- Fixes: #1246

---

- [ ]  Generic LLM v0.0.0

## Summary by CodeRabbit

* **Performance**
  * Improved asset delivery: build assets now use a faster network-first path with timeout and caching, reducing load delays.
  * Static assets are served from cache-first for quicker repeat visits.
* **Reliability / Bug Fixes**
  * General requests now bypass caching, avoiding stale responses and reducing unexpected offline fallbacks.

### DIFF
--- a/src/service-worker.ts
+++ b/src/service-worker.ts
@@ -17,6 +17,8 @@ const CACHE = `cache-${version}`;
 
 const STATIC_HOST = 'https://static.openfoodfacts.org';
 
+const NETWORK_TIMEOUT_MS = 3000;
+
 self.addEventListener('activate', (event) => {
 	// Remove previous cached data from disk
 	async function deleteOldCaches() {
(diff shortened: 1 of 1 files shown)


VaiTon (approved):

VaiTon (changes requested):
On a second look, this is not ok. We already handle the static case on line 41. This was meant to allow looking at products even when the connection is not up.

Maybe it was too pretentious to try to implement this naively via cache and we should just drop the entire thing and just cache static assets.

src/service-worker.ts
```diff
@@ -37,43 +37,55 @@ self.addEventListener('fetch', (event) => {
...
-			// if there's no cache, then just error out
-			// as there is nothing we can do to respond to this request
-			throw err;
+				// if there's no cache, then just error out
+				// as there is nothing we can do to respond to this request
+				throw err;
+			}
 		}
+
+		// 3. Everything else (API calls, etc.) — plain network passthrough, never cached
```
VaiTon: AI comment.

VaiTon (changes requested):

roshinit-a:
@VaiTon pushed another update - added an isSameOriginAsset guard using request.destination to restrict same-origin caching to JS/CSS build artifacts only. Everything else falls through to the plain fetch() passthrough.
Also updated the PR description with a full three-branch flow diagram mapping each request type to its caching behavior.

VaiTon:
>  Generic LLM v0.0.0

@roshinit-a are you sure?

roshinit-a:
> > Generic LLM v0.0.0
> 
> @roshinit-a are you sure?

Yeah I used AI for the comments and description, not great at written English. The fix itself was mine but the second commit was basically me just applying what the bots suggested. I can rework it properly if needed.

VaiTon (approved):

PR #1249, opened 2026-03-23
fix(search): guard against undefined env var and null autocomplete response

### LINKED ISSUE #1248: Bug: Two null-safety issues in `src/lib/api/search.ts` — `getSearchBaseUrl()` misses `undefined` env var, `autocomplete()` can return `undefined` as `AutocompleteResponse`
## Description

Two related null-safety bugs exist in `src/lib/api/search.ts` that can cause hard-to-debug runtime crashes.

---

## Bug 1: `getSearchBaseUrl()` doesn't guard against `undefined`

**Lines 10–17**

```ts
export function getSearchBaseUrl() {
    if (env.PUBLIC_SEARCH_BASE_URL == '') { // ← only catches empty string, not undefined
        throw new Error('PUBLIC_SEARCH_BASE_URL is not set...');
    }
    return env.PUBLIC_SEARCH_BASE_URL; // ← can return undefined if the env var is missing
}
```

SvelteKit types `env.PUBLIC_SEARCH_BASE_URL` as `string | undefined`. The guard only catches the `''` case. If the env var is missing from `.env`, `getSearchBaseUrl()` returns `undefined`, and the immediate caller (`createSearchApi`) does `new URL(undefined)` — which throws an **unreadable `TypeError`** instead of the clear error message already written in the `throw`.

**Fix:** Change `== ''` to `!env.PUBLIC_SEARCH_BASE_URL` to cover both `''` and `undefined`.

---

## Bug 2: `autocomplete()` casts `undefined` to `AutocompleteResponse`

**Line 31**

```ts
export const autocomplete = async (
    query: AutocompleteQuery,
    fetch: typeof window.fetch
): Promise<AutocompleteResponse> => {
    const api = createSearchApi(fetch);
    const response = await api.autocomplete(query);
    return response.data as AutocompleteResponse; // ← no null check
};
```

`response.data` is `undefined` when the search API returns an error or non-200 response. The `as AutocompleteResponse` cast silences TypeScript, but any caller that does `.options.map(...)` on the result will crash with `TypeError: Cannot read properties of undefined (reading 'map')`.

**Fix:** Return a safe fallback: `return (response.data ?? { options: [] }) as AutocompleteResponse;`

---

## Proposed Fix (both bugs together, same file)

```diff
 export function getSearchBaseUrl() {
-    if (env.PUBLIC_SEARCH_BASE_URL == '') {
+    if (!env.PUBLIC_SEARCH_BASE_URL) {
         throw new Error(
             'PUBLIC_SEARCH_BASE_URL is not set. Please set it in your environment variables.'
         );
     }
     return env.PUBLIC_SEARCH_BASE_URL;
 }
```

```diff
-    return response.data as AutocompleteResponse;
+    return (response.data ?? { options: [] }) as AutocompleteResponse;
```

roshinit-a:
### Description

Two related null-safety improvements in `src/lib/api/search.ts`.

**Bug 1 — `getSearchBaseUrl()` doesn't guard against `undefined`**

SvelteKit types `env.PUBLIC_SEARCH_BASE_URL` as `string | undefined`. The previous guard only caught the `''` case. If the env var is missing from `.env`, the function returns `undefined`, and the immediate caller (`createSearchApi`) does `new URL(undefined)` — throwing a cryptic `TypeError` instead of the clear error message already written in the `throw`.

```diff
-    if (env.PUBLIC_SEARCH_BASE_URL == '') {
+    if (env.PUBLIC_SEARCH_BASE_URL == null || env.PUBLIC_SEARCH_BASE_URL == '') {
```

**Bug 2 — `autocomplete()` casts `undefined` to `AutocompleteResponse`**

When the search API returns an error or non-200, `response.data` is `undefined`. The previous `as AutocompleteResponse` cast silenced TypeScript, but any caller doing `.options.map(...)` would crash with `TypeError: Cannot read properties of undefined`.

Per maintainer feedback, the fix is to return the same `{ data, error }` object shape used by other API calls in this codebase, keeping the data layer honest and leaving UI fallback decisions to the caller.

```diff
- ): Promise<AutocompleteResponse> => {
-     const api = createSearchApi(fetch);
-     const response = await api.autocomplete(query);
-     return response.data as AutocompleteResponse;
+ ): Promise<{ data?: AutocompleteResponse; error?: string }> => {
+     try {
+         const api = createSearchApi(fetch);
+         const response = await api.autocomplete(query);
+         if (response.data) return { data: response.data as AutocompleteResponse };
+         return { error: 'No data returned from autocomplete API' };
+     } catch (e: unknown) {
+         const message = e instanceof Error ? e.message : String(e);
+         return { error: message };
+     }
```

`SearchBar.svelte` was updated to unpack `{ data, error }` and handle the error path explicitly.

### Screenshot or video

_No visual changes — both are logic/type-safety fixes._

- Fixes: #1248

---

- [x] I have addressed maintainer feedback regarding architectural concerns.

- [ ]  Generic LLM v0.0.0

## Summary by CodeRabbit

* **Bug Fixes**
  * Autocomplete errors and exceptions are now surfaced and result in empty suggestion lists instead of breaking the UI.
  * Added safe handling when autocomplete returns no data so suggestions gracefully show as empty.
  * Strengthened search base URL validation to catch missing or whitespace-only values earlier.

* **Tests**
  * Added a unit test to ensure whitespace-only search base URLs are treated as not set.

### DIFF
--- a/src/lib/api/search.ts
+++ b/src/lib/api/search.ts
@@ -13,7 +13,7 @@ import { env } from '$env/dynamic/public';
  * @returns The search API base URL string.
  */
 export function getSearchBaseUrl() {
-	if (env.PUBLIC_SEARCH_BASE_URL == null || env.PUBLIC_SEARCH_BASE_URL == '') {
+	if (env.PUBLIC_SEARCH_BASE_URL == null || env.PUBLIC_SEARCH_BASE_URL.trim() === '') {
 		throw new Error(
 			'PUBLIC_SEARCH_BASE_URL is not set. Please set it in your environment variables.'
 		);
--- a/src/lib/api/search.ts
+++ b/src/lib/api/search.ts
@@ -7,8 +7,13 @@ import type { ProductReduced } from './product';
 import { wrapFetchWithCredentials } from './utils';
 import { env } from '$env/dynamic/public';
 
+/**
+ * Returns the base URL for the search API from the PUBLIC_SEARCH_BASE_URL env variable.
+ * @throws {Error} If PUBLIC_SEARCH_BASE_URL is not set or empty.
+ * @returns The search API base URL string.
+ */
 export function getSearchBaseUrl() {
-	if (env.PUBLIC_SEARCH_BASE_URL == '') {
+	if (env.PUBLIC_SEARCH_BASE_URL == null || env.PUBLIC_SEARCH_BASE_URL == '') {
 		throw new Error(
 			'PUBLIC_SEARCH_BASE_URL is not set. Please set it in your environment variables.'
 		);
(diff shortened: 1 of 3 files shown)


VaiTon (changes requested):

VaiTon (approved):

src/lib/api/search.ts
```diff
@@ -22,13 +27,27 @@ export function createSearchApi(fetch: typeof window.fetch): SearchApi {
...
+		const api = createSearchApi(fetch);
+		const response = await api.autocomplete(query);
+		if (response.data) {
+			return { data: response.data as AutocompleteResponse };
+		}
+		return { error: 'No data returned from autocomplete API' };
+	} catch (e: unknown) {
+		const message = e instanceof Error ? e.message : String(e);
+		return { error: message };
+	}
```
VaiTon, line 50: I'm not quite sure we want a try-catch here. In my opinion we can just fetch the data and cast to `{data?: AutocompleteReponse; error?: ...}`.

VaiTon:
@roshinit-a are you still working on this?

VaiTon:
Closing as stale.

Issue #1268, opened 2026-03-26
Static pages iframe suffers from dual scroll, layout overlapping, and state leaks on SPA navigation


MVRU:
### Summary
The `StaticPageIframe` component used to embed external content on `/static` routes is currently experiencing several critical UI and state management issues:
1. **Layout & Scroll:** Users experience a "double vertical scrollbar". The component fails to accurately constrain the iframe's internal overflow, causing it to conflict with the main window's scroll. Additionally, the footer overlaps the content because the container height is not calculated or updated correctly.
2. **SPA Routing State Leak:** When navigating between different `/static` routes, SvelteKit retains the component instance. The `$state` variables (like `frameHeight`) cache the previous page's data, resulting in massive white spaces or clipped content because the height does not update dynamically for the new `src`.
3. **Infinite Loading (Silent Failure):** The component relies entirely on a `postMessage` event from the external source to remove the loading state. If the external site throws a 500 error, a 404, or the user's connection drops, the UI hangs infinitely on "Loading content...".

**Behavior reported by APIPAWE KATOTO Daniel:**
https://openfoodfacts.slack.com/files/U0ANLEUGHMM/F0AMZH4NGLW/screen_recording_2026-03-21_at_21.55.23.mov

**Expected behavior**
- The iframe should seamlessly integrate into the page with only the main window scrollbar visible.
- The footer should sit correctly below the iframe content without overlapping.
- Navigating between static pages should reset the component state and recalculate the height for the new content.
- Network or external server failures should be caught by a timeout, displaying a graceful error state rather than an infinite loader.

**Technical context for the fix**
The solution requires enforcing strict overflow CSS constraints, forcing iframe re-instantiation via Svelte's `{#key}` block to clear the History API and component state, and adding a timeout boundary for the `message` event listener.

harsh-791:
Hi @VaiTon , can I take this up? There hasn’t been any activity on it for a while.

PR #1269, opened 2026-03-26
refactor: overhaul iframe component for stability, routing, and loading states

MVRU:
**Resolves:** #1268 

### Overview
This PR resolves the ongoing layout and state management issues with the `StaticPageIframe` component. The previous implementation suffered from technical debt regarding SvelteKit's SPA lifecycle, lacking proper cleanup and error boundaries.

### Key Architectural & UI Changes:
* **Fixed SPA State Leaks:** Wrapped the `<iframe>` in a `{#key src}` block. This forces Svelte to destroy and recreate the DOM node on route changes, preventing the iframe's internal History API from hijacking the browser's "Back" button and ensuring a clean load state.
* **Resolved Dual Scroll & Footer Overlap:** Enforced `overflow-hidden` and `scrolling="no"` on the iframe. The container now strictly dictates the height based on the parsed `postMessage` payload, preventing the footer from floating over the content.
* **Added Error Boundaries (Timeout):** Implemented an 8-second timeout fallback. If the external source fails to send the `frameHeight` payload (e.g., external server crash, network drop), the UI now aborts the loading state and displays a graceful error message instead of an infinite loader.
* **Security & SSR Improvements:**
  * Made URL parsing SSR-safe with a `try-catch` block to prevent Server 500 crashes if a malformed `src` is provided.
  * Hardened the `message` event listener with strict `origin` validation and payload sanitization (`parseInt`).
* **UX Polish:** Replaced the abrupt conditional rendering (`{#if}`) with a CSS opacity crossfade. This eliminates the visual gap between the loader disappearing and the iframe rendering, mitigating Cumulative Layout Shift (CLS).

### Testing Notes
- Verify navigation between two different `/static` pages (height should recalculate, no cached height).
- Verify scroll behavior (only one main scrollbar should be visible, footer should respect the document flow).
- Simulate a network failure or point `src` to an invalid URL to verify the 8-second timeout fallback triggers correctly.

## Summary by CodeRabbit

* **New Features**
  * Optional title for embedded content (defaults to "External Content")
  * Embedded content fully re-initializes when its source changes and shows a loading overlay until ready

* **Bug Fixes**
  * Height updates are validated before being applied to prevent invalid sizes
  * Visible loading timeout with an 8s error state if no size message arrives
  * Improved cleanup of listeners and timers on unmount or source change

### DIFF
--- a/src/lib/ui/StaticPageIframe.svelte
+++ b/src/lib/ui/StaticPageIframe.svelte
@@ -75,9 +75,9 @@
 	style="height: {frameHeight ? `${frameHeight}px` : '24rem'};"
 >
 	<div
-		class="pointer-events-none absolute inset-0 z-10 flex w-full flex-col items-center justify-center bg-gray-200 transition-opacity duration-300 {showContent
-			? 'opacity-0'
-			: 'opacity-100'}"
+		class="absolute inset-0 z-10 flex w-full flex-col items-center justify-center bg-gray-200 transition-opacity duration-300 {showContent
+			? 'pointer-events-none opacity-0'
+			: 'pointer-events-auto opacity-100'}"
 	>
 		{#if hasTimeoutError}
 			<span class="font-medium text-red-500">Failed to load content.</span>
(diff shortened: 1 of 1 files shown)


VaiTon (changes requested):
Thank you for the PR! I've made some comments

VaiTon:
@MVRU are you still working on this?

PR #1270, opened 2026-03-29
fix: implement graceful degradation for API failures in homepage load

### LINKED ISSUE #1263: Main website crash with 500 (Internal Error)
### What

Production env throws 500 on access.

This may be caused either by:

- The new restriction imposed on /facets/, which must be queried from a logged in client. We query it without credentials to get the contributor count.
  https://world.openfoodfacts.org/facets/contributors.json

- The new permission query we do. I get this in console:
  ```
  Failed to fetch user permissions: HTTP 400
  ```

### Steps to reproduce the behavior:
1. Go to <https://openfoodfacts-explorer.vercel.app/>

adityav27:
## Related issue(s)
Fixes #1263

## The Situation
When external APIs fail (like the current 403 error on the contributors endpoint), the homepage crashes with a 500 Internal Server Error due to unhandled exceptions in the load function.

## The Fix
Implemented graceful error handling at the UI level using Svelte's '{#await}' pattern instead of trying to handle errors in the backend.

**Key Changes:**
- Removed try/catch blocks from 'getNumberOfProducts' and 'getNumberOfContributors'
- Functions now throw errors naturally, which Svelte catches at the template level
- Implemented '{#await}' blocks in '+page.svelte' for both product and contributor counts
- Each async value now has three states:
  - **Pending**: Shows loading skeleton while fetching
  - **Then**: Displays the actual count value
  - **Catch**: Shows fallback value of 0 if API fails
- Maintained parallel fetching pattern for optimal performance

## How It Works
When an API fails:
1. Loading skeleton appears briefly
2. Promise rejects and is caught by '{#await}' catch block
3. Fallback value (0) is displayed gracefully
4. Homepage stays functional instead of showing 500 error

## Why This Approach
- More idiomatic Svelte - follows framework best practices
- Better separation of concerns - UI handles presentation errors
- Better UX - shows loading states and gracefully degrades
- Maintainer recommended - as suggested in PR feedback

## Testing
 TypeScript check: PASSED ('pnpm check')
 Linting: PASSED ('pnpm lint')  
 Formatting: PASSED ('pnpm format')
 Manual testing: Verified homepage stays up when APIs fail
 Graceful degradation: Confirmed fallback values display correctly

### DIFF
--- a/src/routes/+page.svelte
+++ b/src/routes/+page.svelte
@@ -141,7 +141,13 @@
 	>
 		<IconMdiDatabase class="text-primary mb-4 h-12 w-12" />
 		<h2 class="text-xl font-bold">
-			<CountUp value={data.productCount} />
+			{#await data.productCount}
+				<div class="skeleton h-8 w-24"></div>
+			{:then count}
+				<CountUp value={count} />
+			{:catch}
+				<div>0</div>
+			{/await}
 		</h2>
 		<p class="text-base-content/70">{$_('landing.products_count')}</p>
 	</a>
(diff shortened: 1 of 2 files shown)


adityav27:
@VaiTon, Thanks for the feedback! I've updated the implementation to using await.

PR #1289, opened 2026-04-04
fix: debounce folksonomy value lookup and prevent stale updates

pranjal200410:
Replacement for closed #1230.

This PR reapplies the same debounce and stale-response fix on a fresh branch from latest upstream main.

Problem
- Debounce timer state was not preserved across calls.
- Each keystroke could trigger immediate API calls.
- Stale responses could overwrite newer input state.
- Timeout cleanup was missing on effect teardown.

Solution
1. Use a closure-based debounce function with a cancel method.
2. Snapshot requested key and ignore stale responses.
3. Add effect cleanup to cancel pending timers.
4. Guard error handling against stale state.

Validation
- pnpm install --frozen-lockfile
- pnpm lint: pass
- pnpm check: 0 errors, existing warning only
- pnpm build: codespace process exits with 143, reproducible on upstream main in this environment

### DIFF
--- a/src/routes/products/[barcode]/Folksonomy.svelte
+++ b/src/routes/products/[barcode]/Folksonomy.svelte
@@ -99,32 +99,67 @@
 
 	let possibleValues: { v: string; product_count: number }[] | null = $state(null);
 
-	$effect(() => {
-		// when newKey changes, fetch possible values
-		const key = newKey;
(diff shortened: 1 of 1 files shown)


VaiTon:
@pranjal200410 are you still working on this....?

pranjal200410:
yea, I am still working on this

PR #1290, opened 2026-04-04
feat: implement post login redirect

swastikCommits:
### Description
Currently, users are always redirected back to the home page after a successful login through Keycloak. This is a poor UX, especially when a user was in the middle of viewing or searching a product.

So I fixed this by storing the current URL as a "redirect" tag in `localStorage` whenever the user clicks "Login". Once the user successfully logs in, it finds that tag and immediately returns the user to the page where the user left off. 

---

- [ ]  Generic LLM v0.0.0

## Summary by CodeRabbit

* **New Features**
  * Login links now include a redirect parameter and the app preserves the intended destination so users return to their original page after signing in.
  * Post-login flow reads the preserved destination and navigates users back when available; status text now shows a generic "Redirecting..." message.

* **Bug Fixes / Security**
  * Redirect handling enforces same-origin validation and rejects unsafe redirects, falling back to the homepage.

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -63,6 +63,10 @@ export function requireInt(value: string | null, error: () => never) {
  * @returns A safe relative path if valid, or '/' as a fallback
  */
 export function getSafeRedirectUrl(redirectUrl: string | null | undefined, origin: string): string {
+	if (!origin || !URL.canParse(origin)) {
+		throw new Error(`getSafeRedirectUrl: invalid origin "${origin}"`);
+	}
+
 	if (!redirectUrl || !redirectUrl.startsWith('/') || redirectUrl.startsWith('//')) {
 		return '/';
 	}
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -55,3 +55,26 @@ export function requireInt(value: string | null, error: () => never) {
 	}
 	return intValue;
 }
+
+/**
+ * Validates and sanitizes a redirect URL to prevent Open Redirect vulnerabilities.
+ * @param redirectUrl - The URL string to validate
+ * @param origin - The application's current origin (e.g. url.origin)
+ * @returns A safe relative path if valid, or '/' as a fallback
+ */
+export function getSafeRedirectUrl(redirectUrl: string | null | undefined, origin: string): string {
+	if (!redirectUrl || !redirectUrl.startsWith('/') || redirectUrl.startsWith('//')) {
+		return '/';
+	}
+
+	try {
+		const targetUrl = new URL(redirectUrl, origin);
+		if (targetUrl.origin !== origin) {
+			return '/';
+		}
+		// Return only the relative portion to ensure safety
+		return targetUrl.pathname + targetUrl.search + targetUrl.hash;
+	} catch {
+		return '/';
+	}
+}
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -63,7 +63,14 @@ export function requireInt(value: string | null, error: () => never) {
  * @returns A safe relative path if valid, or '/' as a fallback
  */
 export function getSafeRedirectUrl(redirectUrl: string | null | undefined, origin: string): string {
-	if (!origin || !URL.canParse(origin)) {
+	if (!origin) {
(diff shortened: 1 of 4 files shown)


VaiTon:
> Update redirect success message to reflect dynamic destination.
> 
> The message "Redirecting to the homepage..." is now inaccurate since users may be redirected to their previous page.

@swastikCommits (from coderabbitai)

VaiTon (changes requested):
We verify the redirect url two times. Before saving it and before redirecting to it. That's good. What's not good is that the logic for verification is duplicated. Can we extract it to a function and then use the function in both places?

swastikCommits:
Yeah, definitely we can put the logic inside a function and use it in the required places.

src/lib/utils.ts
```diff
@@ -55,3 +55,26 @@ export function requireInt(value: string | null, error: () => never) {
...
+		const targetUrl = new URL(redirectUrl, origin);
+		if (targetUrl.origin !== origin) {
+			return '/';
+		}
+		// Return only the relative portion to ensure safety
+		return targetUrl.pathname + targetUrl.search + targetUrl.hash;
+	} catch {
+		return '/';
+	}
+}
```
VaiTon, line 91: Add something like

```
if (!origin || !URL.canParse(origin)) {
  throw new Error(`getSafeRedirectUrl: invalid origin "${origin}"`);
}
```

src/routes/oauth/login/+page.ts
```diff
@@ -38,6 +39,15 @@ export const load: PageLoad = async ({ url }) => {
...
 	localStorage.setItem('authState', state);
 
+	// Only store if it's a safe, same-origin relative path
+	const redirectUrl = url.searchParams.get('redirect');
+	const safeRedirect = getSafeRedirectUrl(redirectUrl, url.origin);
+	if (safeRedirect !== '/') {
+		localStorage.setItem('authRedirect', safeRedirect);
+	} else {
+		localStorage.removeItem('authRedirect');
+	}
```
VaiTon: Let's always store something, even if it is  `/`

src/routes/+layout.svelte
```diff
@@ -269,7 +269,11 @@
 						>{$_('navbar.logout', { default: 'Logout' })}</a
 					>
 				{:else}
-					<a class="btn btn-outline link" href={resolve('/oauth/login')}
+					<a
+						class="btn btn-outline link"
+						href={resolve('/oauth/login') +
+							'?redirect=' +
+							encodeURIComponent(page.url.pathname + page.url.search)}
```
VaiTon: Extract this to a function

src/routes/+layout.svelte
```diff
@@ -393,7 +397,11 @@
 				>{$_('navbar.logout', { default: 'Logout' })}</a
 			>
 		{:else}
-			<a class="btn btn-outline link" href={resolve('/oauth/login')}
+			<a
+				class="btn btn-outline link"
+				href={resolve('/oauth/login') +
```
VaiTon: Same here

VaiTon (changes requested):

src/lib/utils.ts
```diff
@@ -55,3 +55,30 @@ export function requireInt(value: string | null, error: () => never) {
...
+
+/**
+ * Validates and sanitizes a redirect URL to prevent Open Redirect vulnerabilities.
+ * @param redirectUrl - The URL string to validate
+ * @param origin - The application's current origin (e.g. url.origin)
+ * @returns A safe relative path if valid, or '/' as a fallback
+ */
+export function getSafeRedirectUrl(redirectUrl: string | null | undefined, origin: string): string {
+	if (!origin || !URL.canParse(origin)) {
+		throw new Error(`getSafeRedirectUrl: invalid origin "${origin}"`);
```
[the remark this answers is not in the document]
swastikCommits: @VaiTon what do you think about this?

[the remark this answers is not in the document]
VaiTon: @swastikCommits we can go with the proposed fix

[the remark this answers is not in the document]
swastikCommits: @VaiTon fixed

VaiTon (approved):

PR #1296, opened 2026-04-07
fix: log out redirect uri

### LINKED ISSUE #1295: fix logout redirect error (invalid uri when logging out)
### What
When clicking on the log out button, users are not able to log out successfully and the page is redirecting to a invalid uri.

### Steps to reproduce the behavior:
1.Just log in or sign in
2. Then try to log out

### Expected behavior
User should be able to log out successfully

### Screenshots

Image

Katotodan:
### Description
When clicking on the log out button, users are not able to log out successfully and the page is redirecting to an invalid uri.
This PR aims to fix this

#### Before
Screenshot 2026-04-07 at 14 30 09

#### After (Successfully logged out)
Screenshot 2026-04-07 at 14 37 06

- Fixes:  #1295

- [ ]  Generic LLM v0.0.0

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed logout redirect construction so the app now consistently builds a full callback URL (including origin), ensuring users are reliably returned to the app after signing out.

### DIFF
--- a/src/lib/const.ts
+++ b/src/lib/const.ts
@@ -70,6 +70,8 @@ export const TAXONOMY_URL = (taxo: string, productType?: ProductType) =>
 export const OAUTH_IDP_BASE_URL = PUBLIC_AUTH_BASE_URL;
 export const OAUTH_CLIENT_ID = PUBLIC_AUTH_PKCE_ID;
 export const OAUTH_REDIRECT_URI = (url: URL) => url.origin + resolve('/oauth/login/callback');
+export const OAUTH_LOGOUT_REDIRECT_URI = (url: URL) =>
+	url.origin + resolve('/oauth/logout/callback');
 
 export const KEYCLOAK_REALM = PUBLIC_KEYCLOAK_REALM;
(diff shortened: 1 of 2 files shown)


VaiTon:
Did you test this ? Does this fix the issue? Because reading the code it should not, as `resolve` should already add the right prefix.

Katotodan:
Yes @VaiTon 
I have test it locally and it is now working.
Here is a recorded video showing the change

https://github.com/user-attachments/assets/7b1fdedb-c489-47b6-b018-aafe110ae666

Ramneet04:
This works actually.
I checked `resolve`, It only works when a base path is specified in `svelte.config.js`.
Since we haven’t configured that.
Even if we do it doesn't return a full origin  it only returns a subpath like `/appname`.

VaiTon:
Oh ok I understand now why it does work, it is because we treat the logout url differently from the login url callback. The right solution would be to treat them in the same way (see `const.ts:72`.

```
export const OAUTH_REDIRECT_URI = (url: URL) => url.origin + resolve('/oauth/login/callback');
```

@Katotodan sorry for doubting. Can you make this change?

VaiTon (approved):

PR #1301, opened 2026-04-10
fix: SSR rendering of the product list on the home page.

### LINKED ISSUE #1300: Products not displayed on home page
### Story
On home page, there is a list of products to display after the page rendered.
Those products are controlled by an onMount effect, yet the page is a SSR.
Most of the time the onMount effect is not being called because SSR doesn't call onMount, resulting to the product list not being displayed.

#### Current view
Image

#### Expected view
Image

Katotodan:
### Description
The home page has a list of products that will be displayed on the page mounted.
But the component is an SSR, therefore the onMount effect is disregarded.
This PR changes the onMount logic to load while keeping data fetching asynchronous.

#### Before
Screenshot 2026-04-10 at 14 59 50

#### After
Screenshot 2026-04-10 at 15 00 09

- Fixes: #1300 

---

- [ ]  Generic LLM v0.0.0

### DIFF
--- a/src/routes/+page.svelte
+++ b/src/routes/+page.svelte
@@ -1,12 +1,9 @@
 <script lang="ts">
-	import { onMount } from 'svelte';
 	import { resolve } from '$app/paths';
 
 	import { _ } from '$lib/i18n';
-	import { createRobotoffApi, getProductReducedForCard, getBulkProductAttributes } from '$lib/api';
-	import { deduplicate } from '$lib/utils';
+
 	import { personalizedSearch } from '$lib/stores/preferencesStore';
-	import type { ProductAttributeForScoringGroup } from '$lib/api/product';
 
 	import Logo from '$lib/ui/Logo.svelte';
 	import ProductGrid from '$lib/ui/ProductGrid.svelte';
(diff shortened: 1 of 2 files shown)


VaiTon:
>  But the component is an SSR, therefore the onMount effect is disregarded.

That's not true, SvelteKit will run the onMount when the component is mounted in the DOM. I agree about the issue, but moving loading to `page.ts` is not the solution.

Katotodan:
> > But the component is an SSR, therefore the onMount effect is disregarded.
> 
> That's not true, SvelteKit will run the onMount when the component is mounted in the DOM. I agree about the issue, but moving loading to `page.ts` is not the solution.

I also read about that on the svelte docs ("onMount does not run inside a component that is rendered on the server").
https://svelte.dev/docs/svelte/lifecycle-hooks (second paragraph).
But maybe there is a better solution.

Katotodan:
Hello @VaiTon,
please let me know the appropriated way to solve this issue and I will try my best to solve it :)

VaiTon:
@Katotodan the problem is that the server is giving us a 429 Too Many Requests. We have to switch to the search api for this purpose: 

```
https://world.openfoodfacts.org/api/v2/search?fields=code,product_name&code=3760020507350,3274080005003
```

PR #1322, opened 2026-04-18
fix(compare/shared): validate EAN/UPC barcodes before fetch

pranjal2004838:
## Description

Adds strict barcode format validation in the shared compare loader before product fetch calls are attempted.

- Parses and trims barcodes query values.
- Accepts only valid numeric EAN-8, UPC-A, and EAN-13 codes.
- Verifies checksum digits for each supported format.
- Rejects malformed entries early with HTTP 400.

Why this check exists:
The shared link input is user-controlled. Without format checks, malformed tokens still trigger upstream fetch attempts and extra error logging. This change is a minimal fail-fast guard to avoid unnecessary calls for clearly invalid input.

Fixes # (issue)

N/A

---

## Checklist:

**Author Self-Review:**

- [x] I understand the changes I am proposing and why they are needed.

**LLM Usage Disclosure:**

- [x] If I did use an AI Large Language Model, I have reviewed the generated code and text to ensure its accuracy, security, and relevance to the project context and licensing.

**Triggering Code Review:**

- You can request an AI-powered code review by commenting /gemini review on this PR after it is created.

### DIFF
--- a/src/routes/compare/shared/+page.ts
+++ b/src/routes/compare/shared/+page.ts
@@ -1,5 +1,59 @@
 import OpenFoodFacts, { type Product } from '@openfoodfacts/openfoodfacts-nodejs';
+import { error } from '@sveltejs/kit';
 import type { PageLoad } from './$types';
+
+function hasValidEan13CheckDigit(code: string): boolean {
+	const digits = code.split('').map(Number);
+	const checksum =
+		digits.slice(0, 12).reduce((sum, digit, index) => {
+			return sum + digit * (index % 2 === 0 ? 1 : 3);
+		}, 0) % 10;
+
+	const checkDigit = (10 - checksum) % 10;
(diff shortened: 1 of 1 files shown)


VaiTon:
> This prevents malformed tokens from being forwarded to upstream product APIs and reduces noisy failure amplification from invalid shared links.

I don't really understand how this is an issue. Like, why should we check this?

pranjal2004838:
@VaiTon          Good catch! This is just a simple input hygiene thing, not scope creep. User-controlled input means garbage barcodes still trigger API calls and log noise without this. We just fail fast on obviously bad input—keeps things predictable.

VaiTon:
Yeah but it adds code that already is there on the server. If we want to keep this for making it easier for users we should understand how could a user get to this situation. Can you open an issue so we can discuss this? I'm closing this for now.

PR #1333, opened 2026-04-25
fix(utils): validate FileReader result before extracting base64

devamithra137:
### Description

Prevent fileToBase64 from resolving with undefined when the FileReader result is not in the expected data URL format.

## Problem

The current implementation assumes reader.result always contains a valid data URL and directly extracts the base64 part using split(',')[1].  
If the format is unexpected, this can return undefined while still resolving the promise, leading to silent runtime issues.

## Solution

Add a validation step to ensure the base64 portion exists before resolving.  
If not, the promise is rejected, making the failure explicit.

## Impact

- Avoids silent failures (undefined base64 values)
- Improves reliability of the utility
- No behavior change for valid inputs

## Scope

- Minimal, single-function change
- No unrelated modifications
- Lint passes

### Screenshot or video
NA

- Fixes: 

---

- [x ] I have performed a self-review of my own code (including running it).
- [ x] My changes generate no new warnings or errors (linting, console).

- [ ]  Generic LLM v0.0.0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation when converting files to ensure proper error handling and prevent silent failures during file processing operations.

### DIFF
--- a/src/lib/api/product.ts
+++ b/src/lib/api/product.ts
@@ -597,8 +597,14 @@ export function fileToBase64(file: File): Promise<string> {
 		reader.onload = () => {
 			if (typeof reader.result === 'string') {
 				// Remove the data URL prefix (e.g., "data:image/jpeg;base64,")
-				const base64 = reader.result.split(',')[1];
-				resolve(base64);
+				const parts = reader.result.split(',');
+
+				if (parts.length < 2 || !parts[1]) {
+					reject(new Error('Invalid file format for base64 conversion'));
+					return;
+				}
+
+				resolve(parts[1]);
 			} else {
 				reject(new Error('Failed to convert file to base64'));
 			}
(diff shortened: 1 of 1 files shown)


src/lib/api/product.ts
```diff
@@ -597,8 +597,14 @@ export function fileToBase64(file: File): Promise<string> {
 		reader.onload = () => {
 			if (typeof reader.result === 'string') {
 				// Remove the data URL prefix (e.g., "data:image/jpeg;base64,")
-				const base64 = reader.result.split(',')[1];
-				resolve(base64);
+				const parts = reader.result.split(',');
+
+				if (parts.length < 2 || !parts[1]) {
```
VaiTon: Replace `!parts[1]` with `parts[1].trim() === ''`

devamithra137: > Replace `!parts[1]` with `parts[1].trim() === ''`

updated the validation to handle whitespace-only values using trim(). Thanks

VaiTon (approved):

PR #1334, opened 2026-04-26
fix(static-pages): resolve dual scroll, SPA state leak, and infinite loader on iframe (#1268)

### LINKED ISSUE #1268: Static pages iframe suffers from dual scroll, layout overlapping, and state leaks on SPA navigation

harsh-791:
### Description

Reimplements `StaticPageIframe` to resolve the three bugs reported in #1268:

- **Dual scrollbar** — `frameHeight` now defaults to `null` (was `100vh`, which always overshot the viewport) and the iframe carries `scrolling="no"`, so only the main window scrollbar is visible. Once the embedded page posts `frameHeight`, the container is sized to the exact content height in pixels.
- **SPA state leak between `/static/*` routes** — wrapping `<StaticPageIframe>` in `{#key iframeUrl}` at the route level forces the component to unmount and remount on navigation. State (`frameHeight`, `isLoaded`, `hasError`) resets via `$state` initializers, the timeout restarts, and the iframe DOM element is recreated — which also clears its History API entries (the source of the broken back-button behavior).
- **Infinite loader on silent failure** — added an 8s timeout boundary. If no valid `frameHeight` arrives (network drop, 5xx, malformed `src`), `hasError` flips and the overlay swaps "Loading…" for a translated "Failed to load content." in red.

Hardening also included:

- `postMessage` filter validates origin, that `event.source === iframeEl.contentWindow` (prevents cross-iframe contamination when multiple instances exist), and the payload shape (`parseInt` + `Number.isFinite` + `> 0`).
- `try/catch` around `new URL(src)` so a malformed `src` cannot crash SSR.
- Loader uses `{#if}` (DOM removal) with `role="status"` + `aria-live="polite"`.
- New i18n key `static_iframe.load_failed` (en + it); reuses existing `general.loading`.

This supersedes #1269 — credit to @MVRU for the original investigation and architectural direction (`{#key}` block, 8s timeout boundary, origin validation).

### Screenshot or video

https://github.com/user-attachments/assets/7a1a2fbc-8378-4642-9405-c38935023d8a

- Fixes: #1268

---

## Summary by CodeRabbit

* **New Features**
  * Added a localized loading indicator while embedded content loads.
  * Shows a localized error overlay ("Failed to load content.") if iframe sizing or load fails.
  * Applies a safe fallback height and uses the iframe's measured height only after confirmation.
  * Recreates the embedded view when its URL changes to ensure a clean reload.

* **Bug Fixes**
  * Improved reliability and security of embedded content sizing and load handling.


harsh-791:
Hi @VaiTon ,can you please review it and let me know if any changes are required.

src/lib/i18n/messages/it-IT.json
VaiTon, line 1: We want to manage translations using Crowding. Please revert changes to this file.

VaiTon:
@coderabbitai review

src/lib/ui/StaticPageIframe.svelte
```diff
@@ -1,34 +1,87 @@
...
-			signal: abortController.signal
-		});
+				const h = parseInt(e.data.frameHeight, 10);
+				if (Number.isFinite(h) && h > 0) {
+					frameHeight = h;
+					clearTimeout(timeoutId);
+				}
+			},
+			{ signal: ac.signal }
+		);
```
VaiTon: Please revert the `handler` inlining

src/lib/ui/StaticPageIframe.svelte
```diff
@@ -1,34 +1,87 @@
...
+			aria-live="polite"
+			class="absolute inset-0 flex items-center justify-center bg-gray-100"
+		>
+			{#if hasError}
+				<span class="font-medium text-red-500">{$_('static_iframe.load_failed')}</span>
+			{:else}
+				<span class="animate-pulse font-medium text-gray-500">{$_('general.loading')}</span>
+			{/if}
+		</div>
+	{/if}
```
VaiTon: I don't think we need a loading progress here.

VaiTon (changes requested):

src/lib/ui/StaticPageIframe.svelte
```diff
@@ -1,34 +1,87 @@
...
+	let iframeEl = $state<HTMLIFrameElement | null>(null);
+
+	// URL constructor throws on malformed strings; guard against SSR/runtime crashes.
+	const expectedOrigin: string | null = (() => {
+		try {
+			return new URL(src).origin;
+		} catch {
+			return null;
+		}
+	})();
```
VaiTon: Also I don't understand what this is for

VaiTon (changes requested):

src/lib/ui/StaticPageIframe.svelte
harsh-791: It's for validating where postMessage events come from, used on the
filter line below:

    e.origin !== expectedOrigin

The `<iframe>` content lives on a different origin (world.openfoodfacts.org),
and any window with a reference to ours can post `message` events. Without
this check, any third-party script could spoof a `frameHeight` payload and
resize the container to whatever they want. We pre-compute the expected
origin once from `src` so the filter is a strict equality check.

The try/catch is a safety net: `new URL(...)` throws on malformed strings,
which on SSR would 500 the whole page. If parsing fails we set `expectedOrigin`
to `null`, every postMessage gets rejected, and the 8s timeout surfaces the
error state — same behavior as a network failure, no crash.

Happy to inline a comment to that effect in the code if it's useful.

VaiTon: But still, `src` is a prop, and this is not reactive, so if we first pass an url and then a second one with a different origin this won't work. I mean, it could work since we wrap the component usage with a `{#key}`, but it's cleaner if we just make the prop reactive.

src/lib/ui/StaticPageIframe.svelte
```diff
@@ -1,34 +1,77 @@
...
+			}
 		};
 
-		window.addEventListener('message', handler, {
-			signal: abortController.signal
-		});
+		window.addEventListener('message', handler, { signal: ac.signal });
 
 		return () => {
-			window.removeEventListener('message', handler);
```
VaiTon, line 23: Why don't we remove the event listener anymore?

src/lib/ui/StaticPageIframe.svelte
```diff
@@ -1,34 +1,77 @@
...
-></iframe>
+<div class="relative w-full" style:height={frameHeight !== null ? `${frameHeight}px` : '24rem'}>
+	<!-- scrolling="no" prevents the iframe's internal scrollbar; height is
+	     driven by the postMessage payload from the embedded page. -->
+	<iframe
+		bind:this={iframeEl}
+		{src}
+		title="External Content"
+		scrolling="no"
+		class="absolute inset-0 border-0"
```
VaiTon: I don't really like `absolute` used here. Isn't there any other way of doing this?

VaiTon (changes requested):

src/lib/ui/StaticPageIframe.svelte
harsh-791: Good point — done in <SHA>. Switched to `$derived.by(...)` so
`expectedOrigin` recomputes when `src` changes; the parent's `{#key src}`
still handles resetting the rest of the component state.

src/lib/ui/StaticPageIframe.svelte
harsh-791: The AbortController already does it — when `ac.abort()` runs in the cleanup,
the browser removes any listener registered with `signal: ac.signal`. So an
explicit `removeEventListener` would be redundant.

Docs: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#signal
("The listener will be removed when the given AbortSignal object's abort()
method is called.")

Happy to add it back if you'd prefer the explicit form for clarity.

src/lib/ui/StaticPageIframe.svelte
harsh-791: Good catch — restructured in 766603e. Since the only thing the wrapper was
doing was acting as a positioning context for the error overlay, and the
iframe and error are now mutually exclusive, I dropped the wrapper entirely
and switched to {#if hasError} … {:else} <iframe> … {/if}. No more
`absolute inset-0` anywhere.

The iframe still has an inline `style:height` because the value is dynamic
(postMessage payload) — Tailwind can't generate dynamic px classes — but
that's the only inline style left.

harsh-791:
@VaiTon, the requested changes have been addressed, can you please review them

harsh-791:
@VaiTon any update on this?

VaiTon (approved):

PR #1370, opened 2026-05-16
fix: nutrition information check/uncheck logic

### LINKED ISSUE #1345: Fix nutrition info check/uncheck behaviour
- Correct bug(s) where checking or unchecking nutrition info fields does not update the form correctly
- Make sure user interactions with nutritional data triggers and displays expected changes on the product editing page
- Enhances reliability and clarity for contributors editing nutrition facts

swastikCommits:
### Description

"No nutrition information on the product" toggle has been moved to the top of the component (above the serving size) and now toggling it - hides the serving size, and clears all nutrient data. This ensures that the database doesn't end up with "hidden" nutrition values while the product is flagged as having none.

- Fixes: #1345 

---

- [ ]  Generic LLM v0.0.0

## Summary by CodeRabbit

* **Bug Fixes**
  * Serving size input, its tooltip, and validation alert are now displayed only when nutrition data is enabled, removing unnecessary fields from the editor.
  * Toggling "no nutrition data" updates only that flag and no longer clears other nutrition fields, preventing unintended data loss.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openfoodfacts/openfoodfacts-explorer/pull/1370?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/ui/edit-product-steps/NutritionStep.svelte
+++ b/src/lib/ui/edit-product-steps/NutritionStep.svelte
@@ -102,11 +102,8 @@
 		if (input.checked) {
 			product = {
 				...product,
-				no_nutrition_data: true,
-				nutriments: {} as Nutriments,
-				serving_size: ''
+				no_nutrition_data: true
 			};
-			additionalNutrients = [];
 		} else {
 			product = {
 				...product,
--- a/src/lib/ui/edit-product-steps/NutritionStep.svelte
+++ b/src/lib/ui/edit-product-steps/NutritionStep.svelte
@@ -99,17 +99,10 @@
 	function handleNoNutritionData(event: Event) {
 		const input = event.currentTarget as HTMLInputElement;
 
-		if (input.checked) {
-			product = {
-				...product,
-				no_nutrition_data: true
-			};
-		} else {
-			product = {
-				...product,
-				no_nutrition_data: false
-			};
-		}
+		product = {
+			...product,
+			no_nutrition_data: input.checked
+		};
 	}
 
 	const bySeverity = (a: Issue, b: Issue) => {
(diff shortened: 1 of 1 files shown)


VaiTon:
@swastikCommits can you add a photo of Before/After?

swastikCommits:
Refer: https://github.com/openfoodfacts/openfoodfacts-js/pull/867

swastikCommits:
Before: 
Screenshot 2026-05-20 000124


After:
Screenshot 2026-05-17 084517

VaiTon:
@coderabbitai review

src/lib/ui/edit-product-steps/NutritionStep.svelte
```diff
@@ -99,10 +99,17 @@
...
+			product = {
+				...product,
+				no_nutrition_data: true
+			};
+		} else {
+			product = {
+				...product,
+				no_nutrition_data: false
+			};
+		}
```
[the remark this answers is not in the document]
swastikCommits: Not required, as sever would handle that internally if we send no_nutrition_data as true.

src/lib/ui/edit-product-steps/NutritionStep.svelte
```diff
@@ -99,10 +99,17 @@
 	function handleNoNutritionData(event: Event) {
 		const input = event.currentTarget as HTMLInputElement;
```
VaiTon, line 101: What's the behavior difference between the old code and the new one?

`no_nutrition_data` should keep being `true` in the case `input.checked` is true and `false` otherwise, like in the old code.

swastikCommits: My bad. You're right `no_nutrition_data` could have been handled by the older code in a cleaner way, as both of it just did the same thing.
As I was trying to clean the nutriments and serving size directly from Explorer (which isn't required to be done anymore as server can handle it), so I was using the using the if/else block. So basically its now more of a SDK work to pass just the `no_nutrition_data`

PR #1388, opened 2026-05-20
feat(command-palette): add global keyboard-driven command palette

### LINKED ISSUE #1347: Feature: Add command palette for power users
### Summary
Add a command palette to the OpenFoodFacts Explorer app, enabling power users to quickly trigger navigation, search, and frequently-used features via keyboard shortcuts (like `Cmd+K` or `/`).

### Motivation
A command palette increases productivity and accessibility for experienced users. It streamlines common tasks, fits developer workflows, and is now standard in most modern SaaS admin apps and code tools.

### Desired Capabilities
- Open palette using keyboard shortcut (e.g., `Cmd+K`, `/`)
- Quick navigation to routes (search, settings, product pages, etc.)
- Trigger functions and actions (reload, clear cache, etc.)
- Fuzzy search and autocomplete
- Easy extensibility for adding future actions

### Potential capabilities
- Quick navigation to content pages
- Individual settings search with inline toggling 

### Implementation Notes
- Consider popular libraries such as [cmdk](https://github.com/pacocoursey/cmdk) (MIT, Svelte-compatible)
- Follow Explorer's mobile-first and accessibility best practices
- Should be dismissible and non-intrusive

---
**References:**
- [VS Code Command Palette](https://code.visualstudio.com/docs/getstarted/keybindings#_command-palette)
- [cmdk library](https://github.com/pacocoursey/cmdk)

Namraa310806:
# feat: add keyboard-driven command palette (Ctrl/Cmd+K)

### Description

This PR adds a keyboard-driven command palette to Open Food Facts Explorer. Press `Ctrl+K` on Windows/Linux or `Cmd+K` on macOS from anywhere in the app to open it. Commands are generated from modular provider functions at runtime, not hardcoded in the UI, so the palette stays context-aware and easy to extend. The implementation uses Fuse.js for fuzzy search, a singleton keyboard manager to avoid duplicate DOM listeners, and a native fixed-position overlay that matches the project’s existing modal conventions.

---

### What this PR adds

- `Ctrl/Cmd+K` opens the palette globally; `Escape` or clicking outside closes it
- Fuzzy search over command title, keywords, and description with Fuse.js
- Full keyboard navigation: `↑↓` to move, `Enter` to execute, `Home/End` to jump
- Commands requiring parameters enter a sub-input mode before executing
- Product commands appear only on product pages and stay hidden elsewhere
- Focus returns to the previously focused element on close
- Body scroll locks while the palette is open and restores correctly on close

---

#### Provider architecture

Commands are generated from independent provider functions at runtime:

- `navigationCommands`
- `productCommands`
- `utilityCommands`

Each provider receives route/context information and returns `Command[]`. The registry merges and sorts the results. This is necessary because commands are context-aware. Product commands should only appear on product pages, while navigation and utility commands remain global.

This structure keeps responsibilities isolated and makes the system easier to extend. Adding a new command category requires:

1. creating a new provider file
2. importing it in `index.ts`

No UI changes are required for new command categories.

---

#### Fuse.js

Fuse.js was selected for fuzzy search because it provides:

- weighted field matching
- score-ordered search results
- configurable ranking behavior
- lightweight client-side performance

This improves command palette UX significantly:

- title matches rank higher than keyword matches
- keyword matches rank higher than description matches
- partial matches like `"sett"` correctly surface `"Settings"`

The implementation preserves Fuse.js ranking order directly instead of re-sorting results afterward.

---

#### Fuse index caching

Constructing a Fuse instance builds an internal search index over the command array. Since commands only change when route context changes, rebuilding the index on every keystroke would introduce unnecessary CPU and GC overhead.

To avoid this:

- the Fuse instance is cached
- the index is only rebuilt when the command array reference changes

A `commandIndexMap` is also precomputed to avoid repeated `Array.indexOf()` calls during rendering.

---

#### Native overlay instead of bits-ui Dialog

`bits-ui` was removed from the project in this PR. The attempted `Dialog.Root / Dialog.Portal / Dialog.Content` API does not exist in the version the project had installed and caused runtime issues. The project’s existing modal convention already uses native elements with DaisyUI styling, so this implementation follows the same pattern using:

- a fixed overlay div
- manual focus trapping
- scroll locking via `$effect`

This keeps the implementation lightweight, dependency-free, and consistent with existing project patterns.

---

#### Singleton keyboard manager

One `document.keydown` listener is shared across all registered handlers. Components call `addKeyboardListener` on mount and `removeKeyboardListener` on unmount; they never touch the DOM directly. The teardown function is named `destroyKeyboardManager` to make explicit it is for full application shutdown only, not component cleanup.

Calling it on component unmount would destroy all registered handlers across the app, so that path is intentionally avoided.

---

#### Commands requiring input

Some routes require parameters before navigation.

Example:

- `/search` requires `?q=`

Commands marked with `requiresInput: true` enter sub-input mode on `Enter`: the command list is replaced by a single text field, the user types the required value, and `Enter` calls `command.action(inputValue)`. `Escape` returns to the command list without closing the palette.

---

### Alternatives considered

| Option | Reason rejected |
|---|---|
| `cmdk` | React-only |
| `fuzzysort` | No weighted-key API; performance difference is irrelevant at this scale |
| Per-component keyboard listeners | Duplicate DOM listeners, memory leak risk |
| `bits-ui` Dialog | Version API mismatch, caused runtime crash |
| Debounced search | Adds latency with no benefit at the current command count |

---

### How to add commands

```ts
// src/lib/commands/myFeature.ts
import type { CommandProvider } from './types';

export const myFeatureCommands: CommandProvider = () => [
  {
    id: 'my-command',
    title: 'My Command',
    description: 'Does something useful',
    keywords: ['useful', 'action'],
    category: 'Navigation',
    priority: 50,
    action: () => {
      // command logic
    }
  }
];
```

```ts
// src/lib/commands/index.ts — one new import, one new spread
import { myFeatureCommands } from './myFeature';

export function createCommandRegistry(context = {}) {
  return sortCommands([
    ...navigationCommands(context),
    ...productCommands(context),
    ...utilityCommands(context),
    ...myFeatureCommands(context),
  ]);
}
```

No UI changes are required.

---

### Future extensibility

The current implementation is optimized for the existing command count, but the architecture supports future scaling:

- lazy-loaded providers using dynamic `import()`
- Web Worker-based fuzzy search
- frequently used command ranking
- persisted command history/favorites

These extensions can be added without modifying the UI layer.

---

### Local development note

Routes like:

- `/search`
- `/explore`
- `/products/[barcode]`

call external Open Food Facts APIs during development. Without the required local `.env` configuration, these pages may show a server error after navigation.

The command palette itself correctly:

- performs routing
- executes commands
- updates URLs
- triggers the appropriate route transitions

Any API-related errors originate from external service configuration rather than the command palette implementation itself.

---

### Screenshot / Video

https://github.com/user-attachments/assets/8b16e79c-1cbc-4250-afa9-69ab54dd25d8

---

### Related issue

Fixes: #1347 

---

### Files changed

| File | Change | Purpose |
|---|---|---|
| types.ts | Added | `Command`, `CommandCategory`, `ProductContext`, `CommandProvider` types |
| index.ts | Added | Registry builder, sort, search orchestration |
| navigation.ts | Added | Navigation commands: Home, Search, Explore, Facets, Settings, QR, Contribute, License |
| product.ts | Added | Product commands: Edit, Share, Copy Barcode, product-page only |
| utility.ts | Added | Utility commands such as reload and cache clearing |
| keyboard.ts | Added | Global keyboard listener singleton |
| commandPalette.ts | Added | `commandPaletteOpen` writable store |
| CommandPalette.svelte | Added | Main palette UI: overlay, panel, focus trap, sub-input mode |
| fuzzySearch.ts | Added | Fuse.js wrapper with cached index |
| clipboard.ts | Added | Safe clipboard write with fallback |
| cache.ts | Added | App-owned cache clearing using named keys only |
| routes.ts | Added | Extracts product barcode from current pathname |
| +layout.svelte | Modified | Mounts `<CommandPalette />` at the app root |
| package.json | Modified | Added `fuse.js`; removed unused `bits-ui` dependency |

---

### Validation

```bash
pnpm install
pnpm check
pnpm lint
pnpm build
pnpm dev
```

Manual verification:

- `Ctrl/Cmd+K` opens the palette; browser URL bar does not activate
- Typing filters commands; the best match appears first
- Arrow keys navigate; `Enter` executes; `Escape` closes
- Search Products -> sub-input -> query -> correct URL
- Product page -> contextual commands appear at the top
- Non-product page -> contextual commands are absent
- No console errors

---

- [x] I have performed a self-review of my own code
- [x] I understand the changes I am proposing and why they are needed
- [x] My changes generate no new warnings or runtime errors
- [x] I have verified the feature manually across relevant routes

---

### Large Language Models usage disclosure

- [x] Claude (Anthropic, claude-sonnet-4-6) was used as an architectural advisor and code reviewer. It helped identify bugs from the previous attempt and reviewed design decisions. All code was written, integrated, and tested by me.

## Summary by CodeRabbit

* **New Features**
  * Introduced an in-app command palette (Ctrl/Cmd+K) with full-text search, category grouping, and keyboard navigation.
  * Added navigation commands (Home, Search, Explore, Facets, Settings, Scan QR).
  * Added product-specific commands when a barcode is available: Edit product, Copy barcode, and Share (with graceful fallback behavior).
  * Added utility command: Reload app.
  * Commands are sorted predictably; commands that require input prompt you in a dedicated step.


VaiTon:
@Namraa310806 can you fix CI errors? Thank you!

Namraa310806:
@VaiTon Done.

VaiTon:
Great work on the initial implementation! 

While the current "Centralized" approach (using route matching) works for global navigation, it might become a bottleneck as we add more feature-specific commands.

If we want to add a command in a component (like EditProductForm.svelte) that needs to trigger a local function or access internal $state, the current architecture would require us to "leak" that state globally or complexify the routing logic.

I suggest we switch to a "Push" model (similar to how we handle Toasts). This allows any component to register commands when it mounts and automatically clean them up when it unmounts. This gives commands full access to the component's closure.

Namraa310806:
Thanks for the suggestion — that makes a lot of sense.

The current implementation uses a centralized registry because the initial command set is primarily route-driven (navigation, search, product actions inferred from route context), but I agree that this becomes limiting for feature-specific commands that need access to component-local state or functions.

A push-based registration model similar to the toast system would allow components to register commands on mount and clean them up on unmount, while keeping direct access to their closures and local state.

I'd prefer to keep the current PR focused on delivering the command palette itself and explore the registration-based architecture as a follow-up improvement so the scope of this PR stays manageable.

VaiTon:
> I'd prefer to keep the current PR focused on delivering the command palette itself and explore the registration-based architecture as a follow-up improvement so the scope of this PR stays manageable.

Sure, but modifying the architecture of the base registration system in another PR doesn't make a lot of sense, as the core model of the feature would be changed. I'd much prefer if we lay out the correct architecture now instead of merging something that we're sure we will change in the next PR.

Namraa310806:
Understood, I'll refactor the registration system to the push model in this PR before requesting another review

Namraa310806:
I've refactored the command system to use a push-based registration model.

Commands are now registered by mounted components and cleaned up on unmount through a shared command context, following the same context-based pattern used elsewhere in the project.

Changes include:

* Global navigation commands are registered from `+layout.svelte`
* Product-specific commands are registered from the product page component
* The command palette now consumes registered commands from context instead of building them through the previous route-derived registry flow

This keeps the existing user-facing behavior unchanged while allowing feature-specific components to define commands locally and retain access to component state and closures.

I've verified that:

* Global commands remain available across the app
* Product commands appear on product pages
* Product commands are removed again when navigating away
* Existing keyboard navigation, fuzzy search, and command execution behavior continue to work

I've pushed the refactor and would appreciate another review.

src/lib/utils/fuzzySearch.ts
VaiTon, line 1: The fuse instance should be owned by the CommandPalette component

src/lib/utils/routes.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Extract product context (barcode) from a pathname.
+ * Returns { barcode } when pathname matches /products/:barcode
+ */
+export function getProductContextFromRoute(pathname: string): { barcode: string } | null {
```
VaiTon: No need for this anymore, we can just push the commands when we're in the product +layout.svelte

src/lib/keyboard.ts
```diff
@@ -0,0 +1,84 @@
+import { browser } from '$app/environment';
+
+type KeyHandler = (e: KeyboardEvent) => void;
+
+const listeners = new Set<KeyHandler>();
+let initialized = false;
+let globalListener: ((e: KeyboardEvent) => void) | null = null;
+
+function isInputElement(target: EventTarget | null): target is HTMLElement {
```
VaiTon: Please reuse / extract the analogue function in Shortcuts.svelte

src/lib/keyboard.ts
```diff
@@ -0,0 +1,84 @@
+import { browser } from '$app/environment';
+
+type KeyHandler = (e: KeyboardEvent) => void;
+
+const listeners = new Set<KeyHandler>();
+let initialized = false;
+let globalListener: ((e: KeyboardEvent) => void) | null = null;
```
VaiTon: Please do not use this kind of "static" variables. Instead, do as we already do for the shortcut modal and register this in a component

src/routes/products/[barcode]/+page.svelte
```diff
@@ -100,11 +102,61 @@
 			}
 		});
 
+		const commandCtx = getCommandCtx();
+		const sourceId = `product-${product.code}`;
```
Comment (VaiTon) [inline src/routes/products/[barcode]/+page.svelte]: `product.code` can change, make this reactive

src/routes/products/[barcode]/+page.svelte
```diff
@@ -100,11 +102,61 @@
...
+							return;
+						} catch (error) {
+							if (error instanceof Error && error.name === 'AbortError') return;
+						}
+					}
+
+					await copyTextToClipboard(window.location.href);
+				}
+			}
+		]);
```
Comment (VaiTon) [inline src/routes/products/[barcode]/+page.svelte]: As the product.code can change this has to be put inside an $effect statement

package.json
```diff
@@ -100,7 +99,8 @@
 	"dependencies": {
 		"@openfoodfacts/openfoodfacts-webcomponents": "^1.16.0",
 		"@sentry/sveltekit": "^10.55.0",
-		"@webcomponents/webcomponentsjs": "^2.8.0"
+		"@webcomponents/webcomponentsjs": "^2.8.0",
+		"fuse.js": "^7.3.0"
```
VaiTon: We don't need fuse.js as a dependency, we can keep it as a dev dep

VaiTon (changes requested):

src/lib/utils/cache.ts
VaiTon, line 1: Can we postpone the creation of this commands to another PR?

VaiTon (changes requested):

Namraa310806:
@VaiTon  Just to make sure I understand correctly: are you referring to the product-specific commands (Edit product / Copy barcode / Share product), or the command registration infrastructure itself?

I'm happy to split the feature-specific commands into a follow-up PR if you'd prefer to keep this PR focused on the architecture change.

Namraa310806:
The main goal of this PR is the architectural shift from the pull model to the push model. The product-specific commands were mainly added as a concrete example of how the new registration model would be used.

If you'd prefer to keep this PR focused solely on the architecture, I can remove those commands and keep only the registration infrastructure. We can definitely postpone the product-specific commands to a follow-up PR after the architecture is merged.

Happy to go with whichever approach you prefer.

VaiTon:
@Namraa310806 we can leave the added commands in this PR. I'm just not confortable with the cache-related cmds

PR #1398, opened 2026-05-21
fix(search): return safe fallback from autocomplete on missing data 

### LINKED ISSUE #1248: Bug: Two null-safety issues in `src/lib/api/search.ts` — `getSearchBaseUrl()` misses `undefined` env var, `autocomplete()` can return `undefined` as `AutocompleteResponse`

harsh-791:
### Description

`autocomplete()` in `src/lib/api/search.ts` cast `response.data` to `AutocompleteResponse` without a null check. When the search API returns an error or non-200 response, `response.data` is `undefined`, and the `as AutocompleteResponse` cast silenced TypeScript — so any caller doing `.options.map(...)` crashed with `TypeError: Cannot read properties of undefined (reading 'map')`.

This returns a safe `{ options: [] }` fallback instead:

```ts
return (response.data ?? { options: [] }) as AutocompleteResponse;
```

Note: issue #1248 also reported a null-safety bug in `getSearchBaseUrl()` (the `== ''` guard not catching `undefined`). That one is **already fixed on `main`** — the guard now checks `searchBaseUrl == null || searchBaseUrl === ''` — so no change was needed there.

### Screenshot or video

N/A — internal null-safety fix with no visible UI change.

- Fixes: #1248

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved autocomplete reliability: better handling of missing or unexpected API responses to keep suggestions stable.
  * Aborts in-flight suggestion requests and rejects short/empty queries so results and loading state are accurate.
  * Cleaner debounce and loading behavior to make suggestion updates more responsive and robust during typing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openfoodfacts/openfoodfacts-explorer/pull/1398?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/api/search.ts
+++ b/src/lib/api/search.ts
@@ -1,8 +1,4 @@
-import {
-	SearchApi,
-	type AutocompleteQuery,
-	type Product
-} from '@openfoodfacts/openfoodfacts-nodejs';
+import { SearchApi, type Product } from '@openfoodfacts/openfoodfacts-nodejs';
 import type { ProductReduced } from './product';
 import { wrapFetchWithCredentials } from './utils';
 import { env } from '$env/dynamic/public';
@@ -22,15 +18,6 @@ export function createSearchApi(fetch: typeof window.fetch): SearchApi {
 	const { fetch: wrappedFetch, url } = wrapFetchWithCredentials(fetch, new URL(searchBaseUrl));
 	return new SearchApi(wrappedFetch, { baseUrl: url.toString() });
 }
-
-export const autocomplete = async (
-	query: AutocompleteQuery,
-	fetch: typeof window.fetch
-): Promise<AutocompleteResponse> => {
-	const api = createSearchApi(fetch);
-	const response = await api.autocomplete(query);
-	return (response.data ?? { options: [] }) as AutocompleteResponse;
-};
 
 export type AutocompleteOption = {
 	id: string;
--- a/src/lib/api/search.ts
+++ b/src/lib/api/search.ts
@@ -29,7 +29,7 @@ export const autocomplete = async (
 ): Promise<AutocompleteResponse> => {
 	const api = createSearchApi(fetch);
 	const response = await api.autocomplete(query);
-	return response.data as AutocompleteResponse;
+	return (response.data ?? { options: [] }) as AutocompleteResponse;
 };
 
 export type AutocompleteOption = {
--- a/src/lib/ui/SearchBar.svelte
+++ b/src/lib/ui/SearchBar.svelte
@@ -1,5 +1,9 @@
 <script lang="ts">
-	import { autocomplete, type AutocompleteOption } from '$lib/api/search';
+	import {
+		createSearchApi,
+		type AutocompleteOption,
+		type AutocompleteResponse
+	} from '$lib/api/search';
 	import { _, getBrowserLocale } from '$lib/i18n';
 	import { onDestroy } from 'svelte';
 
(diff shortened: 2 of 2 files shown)


harsh-791:
@VaiTon , can you please review this pr

src/lib/api/search.ts
```diff
@@ -29,7 +29,7 @@ export const autocomplete = async (
 ): Promise<AutocompleteResponse> => {
 	const api = createSearchApi(fetch);
 	const response = await api.autocomplete(query);
-	return response.data as AutocompleteResponse;
+	return (response.data ?? { options: [] }) as AutocompleteResponse;
 };
```
VaiTon: I would just remove this function and call the `autocomplete` method of SearchApi in the caller. With that approach we can also use the `{data,error}`convention

VaiTon (changes requested):

src/lib/api/search.ts
harsh-791: Done, Dropped the autocomplete wrapper (and its now-unused AutocompleteQuery import) from search.ts. The caller in SearchBar.svelte now calls SearchApi.autocomplete directly and uses the { data, error } convention: on error it logs and shows no results, otherwise it reads data.options. No other callers used the wrapper. pnpm run check is green.

VaiTon (approved):

PR #1476, opened 2026-06-05
feat: support to delete language and interactive language addition

### LINKED ISSUE #309: Delete language in product edit page

swastikCommits:
Closes #309

- Implemented the ability to delete secondary languages
- Refactored the "Add Language" section into a compact, toggleable "+" button that expands into an autocomplete search dropdown supporting full keyboard navigation
 
### Screenshots
Screenshot 2026-06-29 084750

## Summary by CodeRabbit

* **New Features**
  * Added an inline autocomplete control to quickly add languages with searchable suggestions and full keyboard support (Arrow keys, Enter, Escape).
  * Enhanced the language editing UI with inline add/delete actions per language, including a disabled delete option for the main language with explanatory messaging.
* **Bug Fixes / Improvements**
  * Improved language deletion safety by requiring confirmation for non-main languages and automatically clearing related language-specific fields to prevent stale data.


VaiTon:
@swastikCommits can you rebase on main?

VaiTon (commented):
I'd prefer if we somehow managed to extract the InputAutocomplete component, because right now LanguageStep.svelte has too many responsibilities

src/lib/ui/edit-product-steps/LanguagesStep.svelte
```diff
@@ -27,26 +30,61 @@
...
+			}
+		}
+		// Trigger Svelte reactivity on product
+		product = { ...product };
+
+		// Switch to main language if the active language was the deleted one
+		if (activeLang === code) {
+			activeLang = product.lang;
+		}
+	}
```
VaiTon, line 68: Instead of modifying `product` multiple times do a copy at the start of the function, edit everything and then assign it back

```
const newProd = ...
...
product = newProd
```

src/lib/ui/edit-product-steps/LanguagesStep.svelte
```diff
@@ -95,72 +133,76 @@
...
+					<input
+						id={`product-name-${code}`}
+						type="text"
+						class="input input-bordered w-full text-sm sm:text-base"
+						bind:value={product[`product_name_${code}`]}
+					/>
+				</div>
+			{/if}
+		{/each}
+	{/if}
```
VaiTon: `{#each}` will execute zero times if the array is empty. No need for the if.

src/lib/ui/InputAutocomplete.svelte
```diff
@@ -0,0 +1,149 @@
...
+			? []
+			: items
+					.filter((item: T) =>
+						searchKeys.some((key: keyof T) => {
+							const val = item[key];
+							return (
+								typeof val === 'string' && val.toLowerCase().includes(searchQuery.toLowerCase())
+							);
+						})
+					)
```
VaiTon: Extract this to a function

src/lib/ui/InputAutocomplete.svelte
```diff
@@ -0,0 +1,149 @@
...
+					.slice(0, 10)
+	);
+
+	$effect(() => {
+		const _ = searchQuery;
+		autoCompleteIndex = -1;
+	});
```
VaiTon, line 53: Add a comment

src/lib/ui/InputAutocomplete.svelte
```diff
@@ -0,0 +1,149 @@
...
+
+	function focusInput(element: HTMLInputElement) {
+		element.focus();
+	}
```
VaiTon: Do we really need this?

src/lib/ui/InputAutocomplete.svelte
```diff
@@ -0,0 +1,149 @@
...
+									e.preventDefault();
+									onselect(item);
+									searchQuery = '';
+									autoCompleteIndex = -1;
+									showInput = false;
+								}}
+							>
+								<span>{item.locale || item.name} ({item.en})</span>
+							</button>
+						</li>
```
VaiTon: Extract to a snippet

VaiTon (changes requested):

src/lib/ui/InputAutocomplete.svelte
swastikCommits: Yes it is required. 
We can actually remove this function and instead we can use the native `autofocus` attribute directly on the input element.

src/lib/ui/edit-product-steps/LanguagesStep.svelte
```diff
@@ -104,27 +146,68 @@
...
+		<div class="alert alert-warning w-full text-sm sm:text-base">
+			{$_('product.edit.no_languages_found')}
+		</div>
+	{:else}
+		{#each Object.keys(product.languages_codes ?? {}) as code (code)}
+			<div
+				role="tab"
+				class="tab flex items-center gap-2 text-xs sm:text-sm {code === activeLang
+					? 'tab-active'
+					: ''}"
```
VaiTon: Use https://svelte.dev/docs/svelte/class#Attributes-Objects-and-arrays

src/lib/ui/InputAutocomplete.svelte
VaiTon: Yes, that's better

VaiTon:
> 101-101: ⚠️ Potential issue | 🟡 Minor | ⚡ Quick win
> 
> Localize remaining hardcoded UI strings in this step.
> 
> aria-label="Info", aria-label="Close", and The main language of the product should use $_(..., { default: ... }) like the rest of this file, so assistive text and labels are translated consistently.
> As per coding guidelines, “All user-facing strings must go through svelte-i18n with a default fallback.”
> 
> Also applies to: 114-114, 133-133

Can we also fix this?

VaiTon:
This now reverts the behavior of having all product names displayed in the "Languages & Names" section

swastikCommits:
> This now reverts the behavior of having all product names displayed in the "Languages & Names" section

Yes, how about we do that in another PR, as currently we can fit the "delete" icon beside each of the individual languages but I'm not sure about using the old way of "Adding a new language" ?

VaiTon (approved):

PR #1552, opened 2026-06-22
feat: implement inline login gate on edit route

### LINKED ISSUE #1360: Implement Login Requirement on /edit Route
**Problem:**
The `/edit` route should require authentication. Currently, unauthenticated users may access the route without being prompted to log in.

**Tasks:**
- Add a login-gate component (`Login Required`) before showing `/edit` form content.
- If not authenticated, show a clear "Login Required" prompt.
- After successful login, redirect to the originally requested `/edit` route.

**Acceptance Criteria:**
- Unauthenticated users are prompted to log in when accessing `/edit`.
- After login, users are directed back to their intended edit route.

swastikCommits:
Closes #1360 

### Screenshot or video

image

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated “Login Required” page for product edit/create, including a “return after login” redirect flow.
* **Localization**
  * Added i18n copy for the login-required experience and editor messaging for unauthenticated scenarios.
* **User Experience**
  * Updated product edit screens to disable submit when appropriate and show the “not logged in” warning only when no user is present in development.
  * When access is required, unauthenticated users are redirected to the “Login Required” page.
* **Chores**
  * Updated the header login link to open as an external link.


VaiTon (commented):
Instead of embedding, why don't we create a generic route that makes the user login and then it is redirected? Something like `/loginrequired` ?

swastikCommits:
> Instead of embedding, why don't we create a generic route that makes the user login and then it is redirected? Something like `/loginrequired` ?

Yeah we can do that instead

swastikCommits:
image
image

cc @teolemon @VaiTon

VaiTon:
@coderabbitai review

PR #1767, opened 2026-08-10
feat(search): add CSV export for search results

### LINKED ISSUE #1747: Download search results as CSV
### Motivation

Researchers and power users need to download filtered search results for offline analysis, reporting, and integration with other tools (e.g., spreadsheets or data pipelines).

### Proposed solution

Add an "Export CSV" control on the search results page that takes the current filter set and generates a CSV. Offer column selection (or use a default set). For very large queries, either paginate exports or generate the CSV on the server and provide a downloadable link when ready. Ensure the export respects Include/Exclude, numeric comparators, AND/OR logic, and ingredient toggles.

### Acceptance criteria

- [ ] Export button is available on search results and respects current filters
- [ ] CSV includes configurable columns and a sensible default (name, brand, nutrition grade, key nutrients, barcodes)
- [ ] Large result sets and pagination are handled (streaming or server-side generation) and user is informed of limits
- [ ] Export is accessible and works on desktop and mobile

### Design / mockup

_No response_

### Alternatives considered

Provide an API endpoint for programmatic exports instead of direct downloads from the UI; or allow limited-page export only.

### Additional context

Related to https://github.com/openfoodfacts/openfoodfacts-explorer/issues/1739

harsh-791:
### Description

Adds an **Export CSV** control on the search results page that downloads products matching the current query/filters (same `q` and sort).

- Default columns: barcode, name, brands, quantity, Nutri-Score, Green-Score, NOVA, and key nutrients
- Paginated client-side fetch with a hard cap of 500 rows; toast when the result set is truncated
- Also updates Search API facet/chart/sort field names (`environmental_score_grade`, `nova_group`) so search loads against the current API

### Screenshot or video

https://github.com/user-attachments/assets/8ec74eab-c808-4f58-885a-dcf1c14e282e

- Fixes: #1747

---

## Summary by CodeRabbit

* **New Features**
  * Added CSV export for search results, including product and nutrition details.
  * Added download controls and localized export status notifications.
  * Exports support up to 500 products, with truncation notices for larger result sets.

* **Bug Fixes**
  * Updated Green-Score sorting to use the current environmental score field.
  * Improved CSV compatibility for special characters, line breaks, and formula-like values.


harsh-791:
@swastikCommits @VaiTon,
Could you please review and suggest if any improvements are required?

harsh-791:
@VaiTon any update on this

swastikCommits:
@harsh-971 please address the coderabbit comments

harsh-791:
@swastikCommits all the coderabbit comments have been addressed

VaiTon:
@coderabbitai review

VaiTon (commented):
Instead of custom-escaping, isn't there an NPM package we can use for this?

harsh-791:
@VaiTon Good point — there’s a solid option for this: csv-stringify. It handles proper CSV escaping, and with escape_formulas: true it also covers spreadsheet formula injection. I’ll switch to that instead of the custom escaping.

harsh-791:
@VaiTon Switched to csv-stringify instead of custom escaping, with escape_formulas: true for spreadsheet formula injection.

src/lib/utils/csv.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * CSV helpers for client-side downloads, backed by csv-stringify.
+ */
+
+import { stringify } from 'csv-stringify/sync';
```
VaiTon: This only works on Node, we should use the browser ESM version

https://csv.js.org/stringify/distributions/browser_esm/

VaiTon (changes requested):

src/lib/utils/csv.ts
VaiTon: Or better, we should either use the NodeJS one or the browser one based on the `BROWSER` variable

harsh-791: Does picking csv-stringify/browser/esm/sync when browser is true and csv-stringify/sync otherwise look good?

src/routes/search/+page.svelte
```diff
@@ -127,6 +132,43 @@
...
 	});
+
+	async function handleExportCsv() {
+		if (isExportingCsv || searchResult.count === 0) return;
+
+		isExportingCsv = true;
+		try {
+			const result = await exportSearchResultsCsv({
+				q: data.query,
+				sortBy: selectedSort.value
```
VaiTon: The export should preserve the raw sort_by URL parameter here instead of using selectedSort.value. selectedSort is derived for the dropdown and falls back to SORT_OPTIONS[0] whenever the URL contains a valid sort that is not in the current UI list—for example the legacy ecoscore_grade field. The search page server still forwards that URL value, so the visible results can be ordered differently from the downloaded CSV. Please pass the current URL sort value (with -unique_scans_n as the default) to keep the export order identical to the current results.

teolemon:
Some merge conflicts on pnpm

src/routes/search/+page.svelte
harsh-791: Does using the raw sort_by URL param (defaulting to -unique_scans_n) look correct for keeping export order in sync with the results?

harsh-791:
@teolemon merge conflicts on pnpm have been resolved.

harsh-791:
@VaiTon @teolemon @swastikCommits 
can someone please review the pr and suggest if any changes are required

VaiTon (commented):
You did not test this again with a browser. It crashes :/ . It tries to create a new Buffer (which is only available on Node). Before asking for a review by us please test your implementation. I'll push a commit and merge.

PR #1851, opened 2026-08-27
fix: preserve missing country data in countries map

### LINKED ISSUE #1850: CountriesMap treats countries missing from facet data as zero-product countries
### What happened?

In `src/routes/facets/[facet]/CountriesMap.svelte`, the `updateMap()` function builds `countryData` by iterating over the entire country taxonomy.

When a country is not present in `facet.tags`, its product count is currently defaulted to `0`:

```js
products: productsByTaxoId.get(id) ?? 0
```

This causes countries with no facet data to be added to `countryData` as if they actually have `0` products.

Later, `getStyle()` checks whether a country exists in `countryData` to decide whether to apply the existing `THEME.noData` styling:

```js
const data = countryData.get(numericId);

if (!data) {
    // no data styling
}
```

Because countries missing from `facet.tags` have already been inserted with `products: 0`, they are treated as data-bearing countries instead of "no data" countries. This prevents the intended `THEME.noData` styling from being applied correctly.

The same issue also affects the choropleth legend. The minimum value is calculated from all entries in `countryData`, including the artificial `0` values for countries that were not present in the facet data. As a result, the legend minimum can incorrectly become `0` even when the actual facet data contains only positive product counts.

The underlying issue is that the code conflates a missing facet value with an explicit zero value. Countries absent from `facet.tags` should remain absent from `countryData`, while an explicit `0` returned by the API should remain a valid value.

### Steps to reproduce

1. Open a facet page that displays the country map.
2. Inspect the facet data and note that `facet.tags` contains the countries returned by the facet API.
3. Compare this with the complete country taxonomy loaded by `getTaxo<Country>('countries', ...)`.
4. In `src/routes/facets/[facet]/CountriesMap.svelte`, trace the `updateMap()` logic.
5. Observe that countries missing from `facet.tags` are still added to `countryData` with `products: 0` because of the `?? 0` fallback.
6. Observe that `getStyle()` uses the absence of a `countryData` entry to apply `THEME.noData`.
7. Observe that the legend minimum is calculated from all `countryData` values, so the inserted zero values can force the minimum to `0`.

### Expected behaviour

Countries that are not present in `facet.tags` should remain absent from `countryData` and should use the existing `THEME.noData` styling.

Only countries for which the facet API provides data should participate in the choropleth data range and legend calculation.

An absent facet entry should not be converted into an artificial `products: 0` value.

### Screenshots or video

Not applicable. The issue is caused by the client-side data processing in `CountriesMap.svelte` and can be identified and verified by tracing the data flow in the code.

### URL or barcode (if applicable)

Not applicable — this issue affects the country facet map rather than an individual product page.

### Browser

Chrome / Chromium

### Additional context

The existing code already contains a `THEME.noData` style and a fallback in `getStyle()` for countries that are not present in `countryData`. This suggests that missing facet data is intended to be represented separately from countries with a product count.

A possible fix is to preserve the distinction between missing and zero values:

```js
const products = productsByTaxoId.get(id);

if (products == null) continue;

countryData.set(numericId, {
    name: entry.name?.en ?? id,
    products
});

### PR DESCRIPTION
**Pull request by Namraa310806**
### Description

This PR fixes an issue where `CountriesMap.svelte` treated countries missing from the facet response as having `0` products.

Previously, the map built `countryData` from the complete country taxonomy and used `?? 0` when a country was not present in `facet.tags`. This conflated missing facet data with an explicit zero product count, causing artificial zero values to affect the choropleth legend and preventing the existing `noData` styling from being applied correctly.

The fix separates taxonomy membership from facet data:

- Countries from the taxonomy continue to be rendered on the map.
- `countryData` only contains countries present in `facet.tags`.
- Explicit `products: 0` values are preserved.
- Countries without facet data use the existing `THEME.noData` styling.
- Empty facet data is handled safely so the legend does not display `Infinity`.
- The country-data transformation was extracted into a small `buildCountryData` helper so the regression tests exercise the production logic directly.

Regression tests cover:
- Countries missing from `facet.tags`
- Explicit zero product counts
- Correct legend min/max values
- Missing-vs-zero data handling
- Empty facet data

### Screenshot or video

Not applicable. This is primarily a data transformation and map rendering logic fix, and the affected behavior is covered by regression tests.

- Fixes: #1850

---

- [ ] I did **not** use an LLM or AI tool to write this PR.
- [x] I used an LLM / AI agent — details below:
  - **Agent / tool name and version:** GPT-5.6 Luna
  - **How it was used:** Used an LLM / AI agent to help debug the issue and create proper regression test cases for the fix. The resulting changes were reviewed before submission.
  - **I have reviewed and take full responsibility for all AI-generated code in this PR.**

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved country map accuracy by displaying only countries with available facet data.
  - Correctly preserves countries with zero products without introducing misleading values.
  - Handles empty or incomplete datasets safely, preventing invalid map ranges and display issues.
  - Improved country name and code resolution for more reliable map labeling.

- **Tests**
  - Added coverage for country data mapping, zero-value counts, missing data, and empty results.

### CODE REVIEW

File: src/routes/facets/[facet]/country-data.test.ts
Comment (VaiTon) [inline src/routes/facets/[facet]/country-data.test.ts:1]: This file should be named `country-data.test.ts`

File: src/routes/facets/[facet]/CountriesMap.test.ts [hunk 1]
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * Regression test for CountriesMap missing-vs-zero data bug.
+ *
+ * The bug: CountriesMap was inserting every taxonomy country into countryData,
+ * defaulting missing facet data to 0. This prevented THEME.noData from being
+ * applied and corrupted the legend with artificial zero values.
+ *
+ * The fix: Only add countries to countryData when they are present in facet.tags.
+ * Use `products == null` to distinguish missing from explicit zero.
+ */
```
Comment (VaiTon) [inline src/routes/facets/[facet]/CountriesMap.test.ts]: No need for this verbose explaination

src/routes/facets/[facet]/country-data.test.ts
```diff
@@ -0,0 +1,158 @@
...
+				country_code_3: { en: 'FRA' },
+				country_code_2: { en: 'FR' }
+			},
+			'en:germany': {
+				name: { en: 'Germany' },
+				country_code_3: { en: 'DEU' },
+				country_code_2: { en: 'DE' }
+			}
+		};
+
```
Comment (VaiTon) [inline src/routes/facets/[facet]/country-data.test.ts:24]: Please remove all these useless comments

VaiTon (changes requested):

VaiTon (approved):

PR #1855, opened 2026-08-29
fix(i18n): add the 59 message keys that are used in code but missing from en.json

MatheusMartinho:
### Description

59 keys are referenced from components but do not exist in `src/lib/i18n/messages/en.json`. Crowdin only sees keys that are present in the source catalogue, so none of these strings can ever be translated: every locale falls back to the English `default` value, and the handful of keys with no `default` at all render the raw key path to the user (for example `settings.dev_warning` and `product.datasources.just_now` show up literally in the UI).

This adds the missing keys. The values are the exact `default` strings already written in the code, so English output is unchanged and no locale loses anything. Copy was written only for the five keys that had no default:

| Key | Value |
| --- | --- |
| `general.unknown_error` | reused the existing default from `StandardError.svelte`: "Something went wrong" |
| `product.datasources.just_now` | "Just now" |
| `product.edit.si_percent_vol` | "% vol" (matches the neighbouring `si_grams`, `si_kilojoules`, ... entries) |
| `settings.dev_warning` | "These settings are intended for moderators and developers. Changing them affects how the site behaves for you." |
| `settings.moderator_mode_help` | "Show moderator only tools and actions across the site." |

Two things had to be fixed along the way:

1. `calculator` was used both as a leaf key (`$_('calculator')` in `+layout.svelte`) and as a namespace (`calculator.clear_all` in `NutritionCalculator.svelte`). Those cannot coexist in one JSON tree, so the label now lives at `calculator.title` and the four call sites were updated.
2. The add-to-comparison toasts in `ProductHeader.svelte` were hardcoded English strings. They now go through the same keys `WcProductCard.svelte` already uses, per the rule in AGENTS.md that every user facing string goes through svelte-i18n.

`loading.message` was deliberately left out. It passes a rotating message as its `default`, so adding a static catalogue entry there would override the rotation.

### Screenshot or video

No visual change in English. The change is that the same strings become translatable, and the five keys listed above stop rendering as raw key paths in non-default cases.

- Fixes: none. This is a bug fix, which AGENTS.md allows without a prior issue. Happy to open one first if you would rather track it.

---

Validation run locally on Node 22, pnpm 11.22.0, all green:

```
pnpm format
pnpm lint     # prettier + eslint, no errors
pnpm check    # 1620 files, 0 errors, 0 warnings
pnpm test     # 10 files, 76 tests passed
pnpm build    # success
```

The list of missing keys was produced by scanning every `$_('...')` call in `src/` and checking it against `en.json`, then each added value was diffed back against the `default` written at the call site to make sure they match exactly. If it would help, I am happy to add that scan as a small Vitest case so this cannot regress.

- [ ] I did **not** use an LLM or AI tool to write this PR.
- [x] I used an LLM / AI agent — details below:
  - **Agent / tool name and version:** Claude (Claude Opus 5), agentic
  - **How it was used:** agentic. It ran the key scan, wrote the catalogue entries and the two component changes, and ran the full validation pipeline.
  - **I have reviewed and take full responsibility for all AI-generated code in this PR.**

## Summary by CodeRabbit

* **New Features**
  * Added English translations for calculator controls, navigation, product details, settings, search, QR features, and other interface elements.
  * Added localized labels for the calculator menu across desktop and mobile navigation.

* **Bug Fixes**
  * Product comparison notifications now support localization, including success, duplicate, and failure messages.

### DIFF
--- a/src/lib/i18n/messages/en.json
+++ b/src/lib/i18n/messages/en.json
@@ -14,7 +14,8 @@
 	"navbar": {
 		"login": "Login",
 		"logout": "Logout",
-		"account": "Account"
+		"account": "Account",
+		"welcome": "User Menu"
 	},
 	"search": {
 		"title": "Search - {query} - Open Food Facts Explorer",
@@ -75,7 +76,18 @@
 		"remove_exclude": "Remove exclude filter",
 		"remove_include_name": "Remove {name} filter",
 		"remove_exclude_name": "Remove exclude {name} filter",
-		"see_all_count": "See all ({count})"
+		"see_all_count": "See all ({count})",
+		"barcode_placeholder": "Barcode (e.g. 1234567890123)",
+		"barcode_validation_title": "Barcode must contain digits only",
+		"enter_barcode_below": "Enter the barcode below.",
+		"export_csv": "Export CSV",
+		"export_csv_error": "Failed to export search results",
+		"export_csv_exporting": "Exporting…",
+		"export_csv_page_hint": "{count} products displayed on this page",
+		"export_csv_success": "Exported {count} products",
+		"filters_sidebar_title": "Search filters",
+		"hide_sidebar": "Hide",
+		"search_placeholder": "Search..."
 	},
 	"auth": {
 		"username": "Username",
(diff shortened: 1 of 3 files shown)


VaiTon (approved):

VaiTon:
Thank you!

